PR #4, opened 2019-04-09
Replaced blocking call to UserManager.FindAsync with await\async

gandrisl:


Aleksanderis:
@gandrisl API build is failing now - https://ci.appveyor.com/project/Simoona/simoona, can you check please?
`DomainServicesValidators were referencing identity nugets as dlls` - I already noticed before, so this part is already fixed in previous pull-request (https://github.com/VismaLietuva/simoona/pull/3).

`All checks have passed` is currently confusing, because there are two separate builds - one for backend, and one for frontend. Backend failed and this pull-request was marked as failing, but then frontend build passed and it marked it as successful. Will need to check how to improve this later. :)

PR #17, opened 2019-04-15
Added markdown support to posts/comments

azygis:
- Added markdown support to posts/comments.
- Partially removed emoji picker so the app works.


Smagumas (approved):
need manual testing in test environment with various browsers

PR #20, opened 2019-04-26
Welcome kudos for new users

ArasB10:
Backend code to add pending kudos for new users when they are confirmed. Keep in mind, currently need to add values manually to database whether welcome kudos are enabled, kudos amount and message. Todo would be create a customization where you can set those field from ui.


src/api/Main/DataLayer/Shrooms.DataLayer/Migrations/201904131312193_AddKudosWelcomeColumns.cs
```diff
@@ -0,0 +1,22 @@
+namespace Shrooms.DataLayer.Migrations
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddKudosWelcomeColumns : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.Organizations", "KudosWelcomeAmount", c => c.Int(nullable: false));
```
Aleksanderis: Is this migration working well on existing DB, having entries in Organizations table?
It looks like new added not nullable column might have issues with existing data.

Aleksanderis (approved):

PR #58, opened 2019-09-27
Bugfix/removing user from default wall causes bugs on frontend

jonaslozys:
Bug: after deleting user from main walls followers list he would not we able to see walls section on sidebar thus unable to discover any walls.

After the fix users sees walls section and is able to follow default wall again:
demo


Aleksanderis:
Default ("official") wall considered to be mandatory - user shouldn't be able to unfollow it, or be not following it in on or another way.
Would it be possible to fix the issue in the way, that user shouldn't need to press "follow" it in the beginning? Otherwise user might be still not following the default wall.

Aleksanderis (commented):
Default ("official") wall considered to be mandatory - user shouldn't be able to unfollow it, or being not following it in one or another way.
Would it be possible to fix the issue in the way, that user shouldn't need to press "follow" it in the beginning? Otherwise user might be still not following the default wall.

jonaslozys:
The only way to unfollow the Official wall is by removing user from the table where followers of walls are listed (as shown in the demo gif). Once the default wall is followed 'Unfollow'  button is disabled.

Aleksanderis:
> The only way to unfollow the Official wall is by removing user from the table where followers of walls are listed (as shown in the demo gif). Once the default wall is followed 'Unfollow' button is disabled.

@jonaslozys I'm not speaking about "unfollowing" possibility, but about the case, that user initially is not following it - by corrupted onboarding or whatever, and this is why such fix was needed.
"Once the default wall is followed 'Unfollow' button is disabled." - this is the issue. That "once" shouldn't be needed, because user might not do it.

jonaslozys:
Alright, I have made some changes so that an API would always return Default wall, marked as 'followed', even if there is no relationship between wall and user saved in database.

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Wall/WallService.cs
```diff
@@ -171,7 +171,7 @@ public async Task<WallDto> WallDetails(int wallId, UserAndOrganizationDTO userOr
                     Name = x.Name,
                     Description = x.Description,
                     Type = x.Type,
-                    IsFollowing = x.Members.Any(m => m.UserId == userOrg.UserId),
+                    IsFollowing = x.Type == WallType.Main ? true : x.Members.Any(m => m.UserId == userOrg.UserId),
```
Aleksanderis, line 174: Could it be: `x.Type == WallType.Main || x.Members.Any(m => m.UserId == userOrg.UserId)`?
Ternary operator doesn't help reading where it's not really neccessary. ;)
(Also there is already `x.Type == WallType.Main || ...` syntax used a bit above in the code.)

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Wall/WallService.cs
```diff
@@ -742,7 +742,7 @@ private async Task<IEnumerable<WallDto>> GetAllOrNotFollowedWalls(UserAndOrganiz
                     Id = w.Id,
                     Name = w.Name,
                     Description = w.Description,
-                    IsFollowing = w.Members.Any(m => m.UserId == userOrg.UserId),
+                    IsFollowing = w.Type == WallType.Main ? true : w.Members.Any(m => m.UserId == userOrg.UserId),
```
Aleksanderis, line 745: Same as above

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Wall/WallService.cs
```diff
@@ -768,6 +768,26 @@ private async Task<IEnumerable<WallDto>> GetUserFollowedWalls(UserAndOrganizatio
...
                 .ToListAsync();
+
+            if (!walls.Any(wall => wall.Type == WallType.Main))
+            {
+                var mainWall = await _wallUsersDbSet
+                    .Include(wall => wall.Wall)
+                    .Where(wall => wall.Wall.Type == WallType.Main && wall.Wall.OrganizationId == userOrg.OrganizationId)
+                    .Select(x => new WallDto
+                    {
+                        Id = x.Wall.Id,
```
Aleksanderis: Looks as duplicated code a bit..
Can't be query above extended, so that "Main" wall would be always added? In that case it would have more sense to select from from _wallDbSet_ instead of _wallUsersDbSet_. In plain SQL it would look like something like this (perhaps it could be better optimized, but I'm more than sure that it's solvable in one query):
```
SELECT DISTINCT w.Name, w.Id, w.Description, w.Type, w.Logo, wm.UserId
FROM Walls w
JOIN WallMembers wm on wm.WallId = w.Id or w.Id = 1
WHERE wm.UserId = @UserId
```

Aleksanderis (changes requested):

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Wall/WallService.cs
```diff
@@ -752,23 +752,32 @@ private async Task<IEnumerable<WallDto>> GetAllOrNotFollowedWalls(UserAndOrganiz
 
         private async Task<IEnumerable<WallDto>> GetUserFollowedWalls(UserAndOrganizationDTO userOrg)
         {
-            var walls = await _wallUsersDbSet
-                .Include(x => x.Wall)
-                .Where(x => x.UserId == userOrg.UserId && x.Wall.OrganizationId == userOrg.OrganizationId)
-                .Where(x => x.Wall.Type == WallType.Main || x.Wall.Type == WallType.UserCreated)
```
Aleksanderis, line 758: `x.Wall.Type == WallType.UserCreated` might be still needed, because it should reduce select amount significantly (my mentioned "select" query was just an example).
There are a lot of "dynamic" walls created, which are event related walls (`WallType.Events`) - i.e. when user joins the event, he joins such wall, and in this case these walls are not needed to be returned.

Aleksanderis (changes requested):

Aleksanderis (approved):

PR #60, opened 2019-09-30
Improvement/kudos send button for mobile

edsdck:
Improvement: kudos submitting button now shows on mobile/tablet.

image


src/webapp/src/client/app/wall/wall.html
```diff
@@ -1,10 +1,13 @@
 <div class="row body-container" has-permissions="['POST_BASIC']">
+    <div class="hidden-lg hidden-md">
```
Aleksanderis, line 2: Hiding it with just CSS - doesn't disable DOM elements completely, they just not visible in the browser. It means, that if some data is being loaded for these elements - it will be still loaded, even if not presented to the user. And then it means, that there will be extra requests to the backend.

Please review and validate network traffic in the browser for extra requests. All Kudos related requests are quite heavy (unless not cached). So it's important to keep it optimized.

Aleksanderis (changes requested):

src/webapp/src/client/app/wall/wall.html
edsdck: I think ~~wall-widget~~ right sidebar has the same problem, because it is always loaded even when it isn't shown. Should I look into it also?

Aleksanderis: > I think wall-widget has the same problem, because it is always loaded even when it isn't shown. Should I look into it also?

Sure, nothing bad if it will be fixed. ;) But it can be considered as separate issue/pull-request (depends how big would be the change).

src/webapp/src/client/app/wall/wall.controller.js
```diff
@@ -3,24 +3,38 @@
 
     angular
         .module('simoonaApp.Wall')
+        .constant('minDisplaySizes', {
+            rightSidebar: 992
```
Aleksanderis: Looks like magic number.. Why 992 and not 994 or 892?
I understand, that value might be coming from some CSS, but in this case concerns would be mixed.
How do you imagine you would maintain such code if CSS/styles would be leaked across whole Javascript code?..

src/webapp/src/client/app/wall/wall.controller.js
```diff
@@ -3,24 +3,38 @@
...
+        vm.windowWidth = $window.innerWidth;
         vm.widgetsInfo = {};
         wallRepository.getWidgetsInfo()
-            .then(function(widgetsInfo) { 
+            .then(function (widgetsInfo) {
                 vm.widgetsInfo = widgetsInfo;
             });
+
+        $window.onresize = function () {
+            $timeout(function () {
```
Aleksanderis: I'm not frontend expert, but "responsive" design relying on timeout functions for me looks like a bit too hardcore..
If you can try finding some better frontend expert and he could approve this logic - I won't be against it. ;)
@Thomasu , maybe you can take a look and give recommendations?

Aleksanderis (changes requested):

src/webapp/src/client/app/wall/wall.controller.js
edsdck: Yes, I understand this, I just haven't solved this yet:/

src/webapp/src/client/app/wall/wall.controller.js
edsdck: Ok, so it didn't work as expected. This code does now:
```javascript
var resizeTimer;
$(window).on('resize', function () {
	clearTimeout(resizeTimer);
	resizeTimer  =  setTimeout(function () {
		vm.windowWidth  =  $window.innerWidth;
		console.log("new size")
	}, 250);
});
```
It uses the timer before setting window width to the variable, so basically it sets the width when user finishes resizing. 

So to make me clear this is what happens when you resizing without timer: 
pvz
Unless it doesn't matter and has no impact on performance. What do you think?

Aleksanderis (approved):

PR #61, opened 2019-09-30
Dismiss notifications when post viewed via direct link

Kristupas13:
Improvement/hide notifications when post viewed via email

- Entered via link notification is removed


Aleksanderis (changes requested):
_"New 'Mark As Seen' button next to post"_ and _"Dismissing all notifications, 'Mark As Seen' button is removed from every post"_ - is not what was originally requested in the issue. _"Entered via link notification is removed"_ part was enough.
Good intentions and new ideas are welcome! However currently these two parts are looking a bit poorly implemented from UX perspective - not nice and badly aligned icon, overall too heavy UI due to too many buttons on the post, etc.
My suggestion is to leave only _"entered via link notification is removed"_ part in this request and move remaining parts to separate pull-request, where it could be developed and improved further.

Aleksanderis (approved):

PR #67, opened 2019-10-01
Add dummy migration to fix failing database updates.

jonaslozys:
Here is the explanation for why database update is failing: https://stackoverflow.com/a/57904403
I manged to replicate the issue and after adding empty migration as suggested in thread linked above, the issue was fixed.


Aleksanderis:
You forgot to commit csproj file with added migration..
After I added it - deployment passed successfully.

PR #70, opened 2019-10-14
Improvement/kudos sending ux

edsdck:
New modal for sending kudos to other users. 

kudossending


src/api/Main/PresentationLayer/Shrooms.API/Controllers/Kudos/KudosController.cs
```diff
@@ -150,6 +151,14 @@ public IEnumerable<KudosTypeViewModel> GetKudosTypes()
             return kudosTypeViewModel;
         }
 
+        [HttpGet]
+        [PermissionAuthorize(Permission = BasicPermissions.Kudos)]
+        public KudosTypeViewModel GetSendType()
```
Aleksanderis: Please keep consistent naming. If it's `_kudosService.GetKudosTypeSend`, then why controller endpoint named in different pattern? Also if it's more natural to say `GetSendType`, then perhaps in service it also could be named - `GetSendKudosType`, and then controller endpoint accordingly.

src/webapp/src/client/resources/en_US/kudos.json
```diff
@@ -20,7 +20,7 @@
   "kudosType": "Type",
   "multiplier": "Multiplier",
   "points": "Points",
-  "kudos.amountOfKudos": "Amount of Kudos",
```
Aleksanderis, line 23: Shouldn't it be changed in other language files as well?

src/webapp/src/client/app/kudos/wall-widget/wall-widget.html
```diff
@@ -2,16 +2,16 @@
...
-            <a ace-kudosify-modal ga-track-event="['Kudos', 'wall big plus', 'opened']" data-test-id="kudosify-button-plus">
+            <a ace-kudosify-modal kudosify-type="{{vm.modalTypes.submit}}" ga-track-event="['Kudos', 'wall big plus', 'opened']" data-test-id="kudosify-button-plus">
                 <span uib-tooltip="{{'kudos.submitKudos' | translate}}" 
                       tooltip-placement="bottom" 
                       tooltip-class="word-break" 
                       class="glyphicon glyphicon-plus kudosify-big-plus-btn" 
                       aria-hidden="true"></span>
             </a>
         </div>
-        <div class="kudos-big-button" feature-flag="kudosSendImproved">
```
Aleksanderis, line 13: If "kudosSendImproved" feature-flag no longer needed, then it should be also removed from `config.js`

Aleksanderis (changes requested):

src/webapp/src/client/resources/en_US/kudos.json
Kristupas13: It already existed in en_US language
image
So just fixed the naming. (It had no references)

Other file already had this translation:
image

Aleksanderis (approved):

PR #71, opened 2019-10-14
Improvement/add lottery entity

jonaslozys:
Added entities and migrations for lottery and lottery participant.


src/api/Main/DataLayer/Shrooms.DataLayer/Migrations/201910031316520_AddLotteryAndLotteryParticipant.cs
```diff
@@ -0,0 +1,64 @@
...
+                .Index(t => t.OrganizationId);
+            
+            CreateTable(
+                "dbo.LotteryParticipants",
+                c => new
+                    {
+                        Id = c.Int(nullable: false, identity: true),
+                        LotteryId = c.Int(nullable: false),
+                        UserId = c.String(maxLength: 128),
+                        Entered = c.DateTime(nullable: false),
```
Aleksanderis: Maybe "Joined" would be a better match?

src/api/Main/DataLayer/Shrooms.EntityModels/Models/Lottery/ImageBaseModel.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shrooms.EntityModels.Models.Lotteries
+{
+    public abstract class ImageBaseModel : BaseModelWithOrg
```
Aleksanderis: What's the purpose of having this `ImageBaseModel`? Seems like it's inherited only once.
If it's needed, then naming is confusing. Following `BaseModel` and `BaseModelWithOrg` naming pattern - it would make sense to name it `BaseModelWithImage`.
However I'm still not convinced there is a need to have such base model.. there shouldn't be so many models with images, or with images stored in such format as they would be stored for lotteries.

Aleksanderis:
Don't forget to update branch to latest, otherwise migrations can be broken again.

Aleksanderis (commented):
I would suggest to drop all current migrations (since they are not merged yet) and re-add everything needed in one migration. Could you do please that?
There is no reason to accumulate so many intermediate migrations..

Aleksanderis (approved):

PR #77, opened 2019-11-06
Feature/lottery widget visualization

Kristupas13:
**Lottery widget and detail modal on large screen:**
widgetDemo

**Lottery widget and detail modal on small screen:**
widgetDemoSmall


src/webapp/src/client/app/lotteries/lotteries-widget/lotteries-widget.directive.js
```diff
@@ -0,0 +1,134 @@
...
+                function filterEndedLotteries(){
+                    angular.forEach(scope.latestLotteries, (lottery) => {
+                        if(moment.utc(lottery.endDate)-moment() < 0)
+                        {
+                            scope.latestLotteries.remove(lottery);
+                        }
+                    });
+                }
+            }
+            function getRemainingTime(input){
```
Kristupas13, line 60: This part of the code is questionable. To return the remaining time is fairly difficult due to different languages (moment-locale package is not installed), so i wrote a helper function which returns translated values. Basically this function returns only unit of time (ex: tomorrow) with no time value or unit of time with time value (ex: 2 days, 3 weeks, etc..). So if anyone has a better idea how to implement remaining time with different languages, I'll be happy to refactor this code.

Aleksanderis (approved):

PR #78, opened 2019-11-07
Feature/lottery kudos refund

edsdck:
Introduced methods in KudosService, which are used in lottery refunding process, and added a few additional operations when calculating total and remaining kudos points. 

Made two new DB migrations: first one adds new default kudos type which is used when refunding lottery and another adds s new field to Lottery entity.

Also added refund page in webapp with ability to retry refund if previous failed.

Lastly with this PR comes @Kristupas13 added functionality, which is an option to export lottery participants to excel file.


src/api/Infrastructure/Shrooms.Infrastructure/ExcelGenerator/ExcelBuilder.cs
```diff
@@ -14,12 +15,24 @@ public ExcelBuilder()
...
+                worksheetBuilder
+                    .WithHeader(headerItems)
+                    .WithRows(rows)
+                    .Build();
+
+            return this;
+        }
+
+        public IExcelBuilder AddNewWorksheet(string sheetName, IEnumerable<IEnumerable<object>> rows)
         {
```
Kristupas13, line 31: Using method overloading I added an option to create worksheet without needing the header.

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Kudos/IKudosService.cs
```diff
@@ -49,12 +50,24 @@ public interface IKudosService
         /// <param name="points">requested points</param>
         void AddKudosLog(AddKudosLogDTO kudosDto, decimal points);
 
+        /// <summary>
+        /// Must not call SaveChanges inside.
```
Aleksanderis: These statements are not looking very promising..
1. What is "inside" is implementation details. Interface is not implementation. So comments doesn't look relevant here in the first place.
2. On the other hand - if such statements are needed, it shows that implemented pattern is confusing and can introduce breaking changes (in case some one misses the comment).

src/api/Main/DataLayer/Shrooms.DataLayer/Migrations/201910221426483_SeedRefundKudos.cs
```diff
@@ -0,0 +1,21 @@
...
+    using System.Data.Entity.Migrations;
+    
+    public partial class SeedRefundKudos : DbMigration
+    {
+        public override void Up()
+        {
+            Sql(
+                "INSERT INTO KudosTypes" +
+                "(Name, Value, Created, Modified, Description, Type, IsActive)" +
+                "VALUES ('Refund', 1, GETDATE(), GETDATE(), 'Kudos, refunded to participants of aborted lottery.', 6, 0)"
```
Aleksanderis: Maybe "Aborted lottery refunded Kudos" would be more clear and shorter?

Aleksanderis (approved):

PR #82, opened 2019-11-12
Improvement/404 page not found redirect

Kristupas13:
- Added redirection to **404 page not found** after inserting wrong url

404OnOff


src/webapp/src/client/resources/en_US/errorCodeMessages.json
```diff
@@ -34,6 +34,9 @@
 	"messageNameAlreadyExists": "Name already exists",
 	"messageShopItemAlreadyExists": "This item already exists",
 	"messageAllFieldsAreRequired": "All fields are required",
-	"userHasPendingKudos" : "User has pending kudos"
+	"userHasPendingKudos" : "User has pending kudos",
+	"oops": "Oops!",
+	"404notFound": "404 Not Found",
+	"requestedPageNotFound": "Sorry, an error has occured, Requested page not found!"
```
Aleksanderis: Seems there is a mistype - should be "occurred" and "Requested" shouldn't be with capital.
On the other hand it would be shorter and still very clear to say just - "Sorry, page not found" ("error occurred" often related to "something went wrong", not 404).
Also if you would look into "access denied" error page - it's a bit more fun and geeky. So maybe an idea to make 404 page in similar way. It can be improved separately of course - current fix to have at least something on 404 is more important).

src/webapp/src/client/resources/lt_LT/errorCodeMessages.json
```diff
@@ -34,5 +34,8 @@
 	"messageNameAlreadyExists": "Toks pavadinimas jau egzistuoja",
 	"messageShopItemAlreadyExists": "Tokia prekė jau egzistuoja",
 	"messageAllFieldsAreRequired": "Visi laukai yra privalomi",
-	"userHasPendingKudos": "Naudotojas turi nepatvirtintų kudosų."
+	"userHasPendingKudos": "Naudotojas turi nepatvirtintų kudosų.",
+	"oops": "Oi!",
+	"404notFound": "404 Puslapis nerastas",
+	"requestedPageNotFound": "Atsiprašome, įvyko klaida, ieškomas puslapis nerastas!"
```
Aleksanderis: Same as in english - "įvyko klaida" part is not needed, since it's often associated with 500 errors.

Aleksanderis (changes requested):

Aleksanderis (approved):

PR #89, opened 2019-11-27
Feature/book covers

jonaslozys:
Add ability to automatically add pictures of book covers from google books API.
Admin only needs to click "Update cover pictures", on request received our API begins background job - requesting google books API info about each books which does not have 'CoverPictureUrl' property.

Demo:
bookCoverDemo


Aleksanderis (changes requested):
Discussed changes are needed

Aleksanderis (approved):

PR #101, opened 2019-12-09
Editable default kudos types and bugfixes 

edsdck:
These changes enable edit feature for default kudos types. Only visibility can be changed, though.

Edit page view for default kudos type:
image

Also changed on/off switch colors resulting in improved UX.
switch

Finally, fixed the security problem which allowed to use disabled kudos types for basic users.


Aleksanderis (changes requested):
All good, but I have a question about "improved UX".
Green color for "On" state looks OK and intuitive, but red color for "Off" state doesn't look appropriate.
Red is often associated with warnings/errors, not to something turned off. Turned off looking grey-ish/disabled - is more intuitive. Don't you agree with that?
Maybe there could be some exceptions, but it seems you changed a style globally.

edsdck:
> 
> 
> All good, but I have a question about "improved UX".
> Green color for "On" state looks OK and intuitive, but red color for "Off" state doesn't look appropriate.
> Red is often associated with warnings/errors, not to something turned off. Turned off looking grey-ish/disabled - is more intuitive. Don't you agree with that?
> Maybe there could be some exceptions, but it seems you changed a style globally.

@Aleksanderis,
Yeah, probably I agree with you. But the problem is that when all fields are disabled and the switch is turned off it kinda looks like you cannot switch it on. What do you think?
image

EDIT: what do you think about white? 
image

Aleksanderis:
> > All good, but I have a question about "improved UX".
> > Green color for "On" state looks OK and intuitive, but red color for "Off" state doesn't look appropriate.
> > Red is often associated with warnings/errors, not to something turned off. Turned off looking grey-ish/disabled - is more intuitive. Don't you agree with that?
> > Maybe there could be some exceptions, but it seems you changed a style globally.
> 
> @Aleksanderis,
> Yeah, probably I agree with you. But the problem is that when all fields are disabled and the switch is turned off it kinda looks like you cannot switch it on. What do you think?
> image
> 
> EDIT: what do you think about white?
> image

If all controls are disabled in the form, but toggle button should be enabled - there is probably another UX issue, which is confusing not just because of colors. I think in Nordic Cool library there is "disabled" state, which should look pretty much clear.

Aleksanderis (approved):

PR #108, opened 2019-12-27
Anonymize deleted users

edsdck:
After deleting an user through a web app all user's sensitive information is removed and profile picture deleted from the storage. Also every time user updates a profile picture the old one is deleted too.

### DIFF
--- a/src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Picture/PictureService.cs
+++ b/src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Picture/PictureService.cs
@@ -24,7 +24,7 @@ namespace Shrooms.Domain.Services.Picture
         public async Task<string> UploadFromImage(Image image, string mimeType, string fileName, int orgId)
         {
             var pictureName = GetNewPictureName(fileName);
-            var tenantPicturesContainer = await GetPictureContainer(orgId);
+            var tenantPicturesContainer = GetPictureContainer(orgId);
 
             await _storage.UploadPicture(image, pictureName, mimeType, tenantPicturesContainer);
 
@@ -34,7 +34,7 @@ namespace Shrooms.Domain.Services.Picture
         public async Task<string> UploadFromStream(Stream stream, string mimeType, string fileName, int orgId)
         {
             var pictureName = GetNewPictureName(fileName);
-            var tenantPicturesContainer = await GetPictureContainer(orgId);
+            var tenantPicturesContainer = GetPictureContainer(orgId);
 
             await _storage.UploadPicture(stream, pictureName, mimeType, tenantPicturesContainer);
 
@@ -43,7 +43,7 @@ namespace Shrooms.Domain.Services.Picture
 
         public async Task RemoveImage(string blobKey, int orgId)
         {
-            var tenantPicturesContainer = await GetPictureContainer(orgId);
+            var tenantPicturesContainer = GetPictureContainer(orgId);
 
             await _storage.RemovePicture(blobKey, tenantPicturesContainer);
         }
@@ -56,9 +56,9 @@ namespace Shrooms.Domain.Services.Picture
             return $"{id}{extension}";
         }
 
-        private async Task<string> GetPictureContainer(int id)
+        private string GetPictureContainer(int id)
         {
-            var organization = await _organizationsDbSet.FirstAsync(x => x.Id == id);
+            var organization = _organizationsDbSet.First(x => x.Id == id);
 
             return organization.ShortName.ToLowerInvariant();
         }
--- a/src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Picture/PictureService.cs
+++ b/src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Picture/PictureService.cs
@@ -41,6 +41,13 @@ namespace Shrooms.Domain.Services.Picture
             return pictureName;
         }
 
+        public async Task RemoveImage(string blobKey, int orgId)
+        {
+            var tenantPicturesContainer = _organizationsDbSet.Where(o => o.Id == orgId).Select(o => o.ShortName).First().ToLowerInvariant();
+
+            await _storage.RemovePicture(blobKey, tenantPicturesContainer);
+        }
+
         private static string GetNewPictureName(string fileName)
         {
             var id = Guid.NewGuid().ToString();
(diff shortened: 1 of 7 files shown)


src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Picture/PictureService.cs
```diff
@@ -41,6 +41,13 @@ public async Task<string> UploadFromStream(Stream stream, string mimeType, strin
             return pictureName;
         }
 
+        public async Task RemoveImage(string blobKey, int orgId)
+        {
+            var tenantPicturesContainer = _organizationsDbSet.Where(o => o.Id == orgId).Select(o => o.ShortName).First().ToLowerInvariant();
```
Aleksanderis: For future improvements: it probably would make sense to move picture container formatting logic to separate reusable method. It's important, because currently this logic is duplicated in Upload and Remove logic. If you decide to change that logic in Upload and will change the format how image is saved - Remove will stop working.

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/UserService/UserService.cs
```diff
@@ -150,10 +153,30 @@ public void Delete(string userToDelete, UserAndOrganizationDTO userOrg)
...
+            user.LastName = string.Empty;
+            user.Bio = string.Empty;
+            user.PhoneNumber = string.Empty;
+            user.UserName = string.Empty;
+            user.FacebookEmail = string.Empty;
+            user.GoogleEmail = string.Empty;
+            user.PictureId = string.Empty;
+            user.BirthDay = DateTime.UtcNow;
+
+            _uow.SaveChanges(userOrg.UserId);
```
Aleksanderis, line 178: Is that SaveChanges really needed here, if yes - then why?
"Final" SaveChanges are done in upper Delete method - isn't it enough?

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/UserService/UserService.cs
```diff
@@ -150,10 +153,30 @@ public void Delete(string userToDelete, UserAndOrganizationDTO userOrg)
...
+        private async Task Anonymize(ApplicationUser user, UserAndOrganizationDTO userOrg)
+        {
+            await _pictureService.RemoveImage(user.PictureId, userOrg.OrganizationId);
+
+            user.Email = string.Empty;
+            user.FirstName = string.Empty;
+            user.LastName = string.Empty;
+            user.Bio = string.Empty;
+            user.PhoneNumber = string.Empty;
+            user.UserName = string.Empty;
```
Aleksanderis: Are you sure it will conform all constraints? Like "unique" constraint, "not null/empty", etc?

Aleksanderis (changes requested):

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/UserService/UserService.cs
edsdck: Yes, it is needed, because when soft deleting entity's state is set to _detached_ which means all unsaved changes to an entity are lost.

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/UserService/UserService.cs
edsdck: Yes, you are right. There was a constraint on _Email_ field. I've fixed that now
.

Aleksanderis: > Yes, you are right. There was a constraint on _Email_ field. I've fixed that now
> .

Only email? What about UserName?

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Picture/PictureService.cs
```diff
@@ -34,19 +34,33 @@ public async Task<string> UploadFromImage(Image image, string mimeType, string f
...
         {
             var id = Guid.NewGuid().ToString();
             var extension = Path.GetExtension(fileName).ToLowerInvariant();
 
             return $"{id}{extension}";
         }
+
+        private async Task<string> GetPictureContainer(int id)
+        {
+            var organization = await _organizationsDbSet.FirstAsync(x => x.Id == id);
```
Aleksanderis: Before refactoring that DB call wasn't async.
Making it async might have additional side effects, since all other DB calls here are not async.
Usually it doesn't work well with EntityFramework, when you mix sync and async DB calls..
Please test it carefully - with some different DB operations done concurrently.

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/UserService/UserService.cs
Aleksanderis: Only email? What about UserName?

Aleksanderis (approved):

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/UserService/UserService.cs
edsdck: Yes, only email.

Aleksanderis: I still would suggest to generate some random string for all fields with not-null constraint in DB. Or at least for such important fields like _username_, emails (all emails), etc. In other words - anonymization shouldn't break the user, it just should make it impossible to identify that user.

edsdck: Only _username_ field doesn't allow nulls. I'll fix this then. I could generate 1 Guid and set it to all different fields, what do you think?

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Picture/PictureService.cs
edsdck: I tested it a bit and seems it works as it should, but I'm not sure if I've used a correct approach to do it.
I could make it synchronous if you wish.

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/UserService/UserService.cs
Aleksanderis: Should be fine, I think.

src/api/Main/BuisnessLayer/Shrooms.Domain/Services/Picture/PictureService.cs
Aleksanderis: I would say - let's leave DB call synchronous.
It won't have huge improvement if we have one async call and all other calls still synchronous, but there could be issues.

Aleksanderis (changes requested):

Aleksanderis (approved):

PR #109, opened 2019-12-30
Bugfix/event invitation ical

Kristupas13:
- Added ability to attach file to mail message
- Added iCal nugget
- Added 'Add to calendar' redirect button to google calendar with selected event


Aleksanderis (changes requested):
As discussed - "Add to calendar" button could be extended to a dropdown with two options:
1. Redirect to Google Calendar inside browser
2. Download iCal file with the event

Aleksanderis (approved):

PR #133, opened 2020-02-19
fix infinite scrolling on smartphones

edsdck:


src/webapp/src/client/app/events/list/by-type/by-type.html
```diff
@@ -1,28 +1,29 @@
-<div class="event-list-item-container" 
-     infinite-scroll="vm.addMoreEvents()" 
-     infinite-scroll-distance="0"
-     infinite-scroll-disabled="vm.isEventsLoading || !vm.isScrollingEnabled">
+<div class="event-list-item-container" infinite-scroll="vm.addMoreEvents()" infinite-scroll-distance="1"
```
Aleksanderis, line 1: Comment for future - please avoid changing whitespaces/linebreaks formatting (and/or configure your IDE accordingly).
I also like the style more how you fixed it, but in that case it would be much more changes needed in many files to keep consistency. It also doesn't help to review changes, when it's mailny whitespaces/linebreaks are changed.

Aleksanderis (approved):

src/webapp/src/client/app/events/list/by-type/by-type.html
edsdck: I'll keep that in mind.

PR #147, opened 2020-09-08
Add more space for book comments

Belkas:
A minor visual fix, see https://i.imgur.com/cmrPhAx.png


Aleksanderis:
Seems like it was fixed in a bit wrong way.. Class applied on a wrong tag.
image

PR #195, opened 2022-02-06
Fixed markdown formatting

jusrus01:
Instead of truncating text after a certain number of symbols, I made use of this CSS, where we specify the number of lines  (```-webkit-line-clamp: 21;```) to show in a given container.
```
.show-more-message-container {
    display: -webkit-box;
    -webkit-box-orient: vertical;
    -webkit-line-clamp: 21;
    overflow: hidden;
    text-overflow: ellipsis;
  }
```

### DIFF
--- a/src/webapp/src/client/app/common/directives/wall/message-show-more/message-show-more.directive.js
+++ b/src/webapp/src/client/app/common/directives/wall/message-show-more/message-show-more.directive.js
@@ -1,28 +1,28 @@
 (function () {
-    "use strict";
+    'use strict';
 
     angular
-        .module("simoonaApp.Common")
-        .directive("aceMessageShowMore", messageShowMore);
+        .module('simoonaApp.Common')
+        .directive('aceMessageShowMore', messageShowMore);
 
     messageShowMore.$inject = [
-        "$compile",
-        "wallService",
-        "$timeout",
-        "$window",
+        '$compile',
+        'wallService',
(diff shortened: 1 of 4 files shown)


src/webapp/src/client/app/common/directives/wall/message-show-more/message-show-more.directive.js
```diff
@@ -1,86 +1,133 @@
...
             scope: {
-                message: '@',
-                showMoreType: '@',
-                hasHashtagify: '='
+                message: "@",
+                showMoreType: "@",
+                hasHashtagify: "=",
             },
-            link: linkFunc
+            link: linkFunc,
```
Aleksanderis: Trailing commas also usually shouldn't be added.

src/webapp/src/client/app/common/directives/wall/message-show-more/message-show-more.directive.js
```diff
@@ -1,86 +1,133 @@
 (function () {
-    'use strict';
+    "use strict";
```
Aleksanderis: Isn't it single quotes should be a default? Maybe in Simoona project it's not specifically configured, but usually it's a standard default.
If such things are not clear - for open-source projects usual practice is to leave such things alone and not change them.

Aleksanderis (approved):

PR #212, opened 2022-07-05
Added non-public event search

jusrus01:
Added a separate page for searching events that are not started yet that can be accessed through here (requires EVENT_ADMINISTRATION permission):
image

Features include:
- Searching

- Sorting by columns

- Ability to filter events by multiple event types and/or office types (**not selecting any types to filter by returns everything**)

- Ability to select predefined presets that automatically selects required event/office (or other) types

- Ability to do CRUD operations on presets.

- All of the above also apply to viewing (except searching) details about selected event

In addition to this, some refactoring was done, such as adding additional view models for EventController and creating interfaces for models that can be used for pagination and custom sorting. Furthermore, I made sure that UI layout works well for mobile (375px width support).


src/api/Shrooms.Domain.ServiceValidators/Validators/FilterPresets/IFilterPresetValidator.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Shrooms.Domain.ServiceValidators.Validators.FilterPresets
+{
+    public interface IFilterPresetValidator
+    {
+        void CheckIfFilterTypesAreValid(FilterType[] filterTypes);
+
+        void CheckIfFilterTypesContainsDuplicates(FilterType[] filterTypes);
```
Aleksanderis: Types (plural) -> so it should be "ContainDuplicates" (not Contains).

```suggestion
        void CheckIfFilterTypesContainDuplicates(FilterType[] filterTypes);
```

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                throw new ValidationException($"Entity has to have at least one property");
+            }
+
+            return query.OrderBy(string.Format("{0} {1}", firstProperty.Name, sortDirection));
+        }
+
+        private static bool EntityHasProperty<TEntity>(string propertyName) where TEntity : class
+        {
+            var propertyNames = propertyName.Split('.');
```
Aleksanderis: How come you split one propertyName and get multiple property names? 🤔
Is it a naming issue or it works in some smarter way?

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using System.Linq.Dynamic;
+using System.Reflection;
+
+namespace Shrooms.Domain.Extensions
+{
+    public static class DbSetExtensions
+    {
+        private const string DefaultSortDirection = "desc";
```
Aleksanderis: Why default is "desc"? It's more natural to have "asc" by default.

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+            var firstProperty = typeof(TEntity)
+                    .GetProperties(BindingFlags.Public | BindingFlags.Instance)
+                    .FirstOrDefault();
+
+            if (firstProperty == null)
+            {
+                throw new ValidationException($"Entity has to have at least one property");
+            }
+
+            return query.OrderBy(string.Format("{0} {1}", firstProperty.Name, sortDirection));
```
Aleksanderis: Same thing about string interpolation.

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        private static IQueryable<TEntity> OrderByFirstPropertyName<TEntity>(this IQueryable<TEntity> query, string sortDirection)
+        {
+            var firstProperty = typeof(TEntity)
+                    .GetProperties(BindingFlags.Public | BindingFlags.Instance)
+                    .FirstOrDefault();
+
+            if (firstProperty == null)
+            {
+                throw new ValidationException($"Entity has to have at least one property");
```
Aleksanderis: Are there any entities without properties? Looks a bit too much defensive. If there are entities without properties - I would guess it would fail much earlier, isn't so?

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                sortDirection = DefaultSortDirection;
+            }
+
+            if (propertyName == null || !EntityHasProperty<TEntity>(propertyName))
+            {
+                return query.OrderByFirstPropertyName(sortDirection);
+            }
+
+            return query.OrderBy(string.Format("{0} {1}", propertyName, sortDirection));
```
Aleksanderis: Why not to use string interpolation ("$") instead of string.Format() here?

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+            sortDirection = sortDirection?.ToLower();
+            
+            if (sortDirection == null || (sortDirection != "asc" && sortDirection != "desc"))
+            {
+                sortDirection = DefaultSortDirection;
+            }
+
+            if (propertyName == null || !EntityHasProperty<TEntity>(propertyName))
+            {
+                return query.OrderByFirstPropertyName(sortDirection);
```
Aleksanderis, line 37: Why to force it to sort by first property, SQL would just use some default sort - so it could return just an unmodified query.

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            return query.OrderBy(string.Format("{0} {1}", firstProperty.Name, sortDirection));
+        }
+
+        private static bool EntityHasProperty<TEntity>(string propertyName) where TEntity : class
+        {
+            var propertyNames = propertyName.Split('.');
+            var bindingFlags = BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance;
+
+            var type = typeof(TEntity).GetProperty(propertyNames[0], bindingFlags);
```
Aleksanderis: It's duplicated code, same as in for loop below.
It's either could be structurized in a smarter way, so that a check for the first element would be done in same loop. Otherwise, this check logic should be extracted to a separate method.

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            if (type == null)
+            {
+                return false;
+            }
+            
+            if (propertyNames.Length == 1)
```
Aleksanderis: This looks obsolete. If this check would be skipped, then for loop would be executed just once - so what this separate check gives you? If it's dropped, then it would help to simplify what is stated in the previous comment.

src/api/Shrooms.Domain/Services/FilterPresets/FilterPresetService.cs
```diff
@@ -0,0 +1,259 @@
...
+
+            var filtersDtos = new List<FiltersDto>();
+
+            foreach (var type in filterTypes)
+            {
+                var filters = await GetFiltersQueryByFilterType(type).ToListAsync();
+                var filtersDto = new FiltersDto
+                {
+                    FilterType = type,
+                    Filters = filters,
```
Aleksanderis:
```suggestion
                    Filters = filters
```

src/api/Shrooms.Domain/Services/FilterPresets/FilterPresetService.cs
```diff
@@ -0,0 +1,259 @@
...
+                };
+
+                filtersDtos.Add(filtersDto);
+            }
+
+            return filtersDtos;
+        }
+
+        private bool NewDefaultPresetWasSet(IEnumerable<FilterPreset> createdPresets, IEnumerable<FilterPresetDto> updatedPresets)
```
Aleksanderis: Sounds more natural:
```suggestion
        private bool IsNewDefaultPresetSet(IEnumerable<FilterPreset> createdPresets, IEnumerable<FilterPresetDto> updatedPresets)
```

src/api/Shrooms.Domain/Services/FilterPresets/FilterPresetService.cs
```diff
@@ -0,0 +1,259 @@
...
+                return Enumerable.Empty<FilterPresetDto>();
+            }
+
+            var presets = await _filterPresetDbSet
+                .Where(preset => 
+                    updateDto.PresetsToDelete.Contains(preset.Id) &&
+                    preset.OrganizationId == updateDto.UserOrg.OrganizationId)
+                .ToListAsync();
+
+            var removedPresets = presets.Select(preset => new FilterPresetDto
```
Aleksanderis, line 169: This probably could be moved inside select statement above, before `.ToListAsync()`.
Also it's confusing, that `Delete()` method parameter is called `updateDto`.

src/api/Shrooms.Contracts/DataTransferObjects/FilterPresets/AddEditDeleteFilterPresetDto.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Shrooms.Contracts.Enums;
+using System.Collections.Generic;
+
+namespace Shrooms.Contracts.DataTransferObjects.FilterPresets
+{
+    public class AddEditDeleteFilterPresetDto
```
Aleksanderis: Name sounds like it's really not driven by "single-responsibility" principles.
If it's some batch request on purpose - then to avoid the confusion perhaps it could be named somehow differently (something like "ManageFilterPresetDto").

src/api/Shrooms.Domain/Services/FilterPresets/FilterPresetService.cs
```diff
@@ -0,0 +1,259 @@
...
+
+namespace Shrooms.Domain.Services.FilterPresets
+{
+    public class FilterPresetService : IFilterPresetService
+    {
+        private static readonly SemaphoreSlim _filterPresetUpdateCreateLock = new SemaphoreSlim(1, 1);
+
+        private readonly IUnitOfWork2 _uow;
+
+        private readonly DbSet<FilterPreset> _filterPresetDbSet;
```
Aleksanderis, line 24: Maybe it's by design and copied from some other places - but just doublechecking if is it valid, that it's `DbSet` and not `IDbSet` like below.

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
jusrus01: If we want, for instance, use this method on the Event class and sort by the ICollection<EventParticipant>.Count, then we would need to pass in string like: "EventParticipants.Count". In order to validate that the property exists I need to retrieve ICollection<EventParticipants> property first (also see if it exists in the class) and only then I can view the Count property and validate that it exists.

src/api/Shrooms.Domain/Services/FilterPresets/FilterPresetService.cs
jusrus01: IDbSet does not contain RemoveRange() method. Also, there is a mixed usage of them (some services use both IDbSet and DbSet), so I use DbSet when I use AddRange() or RemoveRange().

src/api/Shrooms.Contracts/DataTransferObjects/FilterPresets/AddEditDeleteFilterPresetDto.cs
jusrus01: We have AddEditDeleteExternalLinkDto for external links, should I rename this too?

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
Aleksanderis: Then it's a naming issue. :)
Perhaps rename `propertyNames` to `propertyNameParts`, or smth?

src/api/Shrooms.Contracts/DataTransferObjects/FilterPresets/AddEditDeleteFilterPresetDto.cs
Aleksanderis: Wasn't it added not so long ago as well?
For me "addeditdelete" sounds really confusing. I would recommend to change it, but of course it's more like a recommendation.

src/api/Shrooms.Domain/Services/FilterPresets/FilterPresetService.cs
Aleksanderis: We discussed it briefly and agreed to leave it as it is.
(In fact I just thought that `presets` and `removedPresets` are same objects.)

src/api/Shrooms.Domain.ServiceValidators/Validators/FilterPresets/IFilterPresetValidator.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        void CheckIfPageTypeExists(PageType page);
+
+        void CheckIfCountsAreEqual<T1, T2>(T1 firstCollection, T2 secondCollection)
+            where T1 : ICollection
+            where T2 : ICollection;
+
+        void CheckIfMoreThanOneDefaultPresetExists(ManageFilterPresetDto manageFilterPresetDto);
+
+        void CheckIfFilterPresetsContainsUniqueNames(IEnumerable<FilterPresetDto> filterPresetDtos);
```
Aleksanderis: Didn't notice it at first - but it same issue with plural and "contains" (should "presets contain").

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
Aleksanderis: This comment seems wasn't addressed. Do you want to leave it as it is ar just missed it?

PR #214, opened 2022-07-07
Fixed lotteries not updating status to "Ended"

jusrus01:
Added another LotteryStatus enum value **Ended** (also renamed previously existing Ended value to Finished based on trello suggestion) to indicate that the lottery ended, because of time and not user action. Updated the UI to allow Abort/Finish the lottery that was not ended by the user. Also, will need to create a logic app in azure for UpdateLotteriesStatus end-point for this fix to work properly.


src/api/Shrooms.Contracts/Enums/LotteryStatus.cs
```diff
@@ -5,9 +5,10 @@ public enum LotteryStatus
         Drafted = 1,
         Started,
         Deleted,
-        Ended,
+        Finished, // User finished the lottery
```
Aleksanderis: I would say it's still confusing - "ended" and "finished" are synonyms, and you still need to have comments to explain it. :)
"Time ran out" - I would say it could be "Expired". Perhaps it would a bit negative meaning (that it finished unintentionally), but at least it would be less confusing, than having both "ended" and "finished".

(I know you did it like suggested in the backlog issue, but perhaps it was needing a better grooming.)

src/api/Shrooms.Premium/Domain/Services/WebHookCallbacks/Lotteries/ILotteryStatusChangeService.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace Shrooms.Premium.Domain.Services.WebHookCallbacks.Lotteries
+{
+    public interface ILotteryStatusChangeService
+    {
+        Task UpdateStartedLotteriesToEndedAsync();
```
Aleksanderis:
```suggestion
        Task ProcessExpiredLotteriesAsync();
```

src/api/Shrooms.Presentation.Api/Web.config
```diff
@@ -75,7 +75,8 @@
 
     <!-- Endpoints for scheduled jobs -->
     <add key="RecurringEventEndpoint" value="/externalpremiumjobs/updaterecurringevents" />
-    <add key="BirthdayNotificationEndpoint" value="/externaljobs/SendBirthdaysNotifications" />
+	<add key="UpdateLotteriesStatus" value="/externalpremiumjobs/UpdateLotteriesStatus" />
```
Aleksanderis: It seems indentation (or whitespaces) is ruined here a bit.

src/webapp/src/client/app/lotteries/lotteries.module.js
```diff
@@ -15,7 +15,8 @@
             ended: 4,
             refundStarted: 5,
             refundLogsCreated: 6,
-            refunded: 7
+            refunded: 7,
+            endedByTime: 8
```
Aleksanderis: It's not a good idea to name enums differently than in the backend.

For fully decoupled frontend and API - it wouldn't be a big problem, but in this case there is is just a frontend and backend-for-frontend.

Aleksanderis (approved):

PR #218, opened 2022-07-12
Improving sorting to event reports

jusrus01:
Created ISortableProperties interface that I use to define models that can be used to sort by multiple properties. From front-end incoming string should look like: **"FirstName asc;LastName desc;"**

### DIFF
--- a/src/api/Shrooms.Contracts/Infrastructure/ISortableProperties.cs
+++ b/src/api/Shrooms.Contracts/Infrastructure/ISortableProperties.cs
@@ -1,0 +1,7 @@
+﻿namespace Shrooms.Contracts.Infrastructure
+{
+    public interface ISortableProperties
+    {
+        string SortByProperties { get; set; }
+    }
+}
--- a/src/api/Shrooms.Contracts/Shrooms.Contracts.csproj
+++ b/src/api/Shrooms.Contracts/Shrooms.Contracts.csproj
@@ -132,7 +132,7 @@
     <Compile Include="Infrastructure\IExcelBuilder.cs" />
     <Compile Include="Infrastructure\IJobScheduler.cs" />
     <Compile Include="Infrastructure\ILogger.cs" />
-    <Compile Include="Infrastructure\ISortableProperties.cs" />
+    <Compile Include="Infrastructure\ISortable.cs" />
     <Compile Include="Infrastructure\ISystemClock.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="ViewModels\AbstractViewModel.cs" />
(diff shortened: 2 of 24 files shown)


src/api/Shrooms.Contracts/Shrooms.Contracts.csproj
```diff
@@ -132,6 +132,7 @@
     <Compile Include="Infrastructure\IExcelBuilder.cs" />
     <Compile Include="Infrastructure\IJobScheduler.cs" />
     <Compile Include="Infrastructure\ILogger.cs" />
+    <Compile Include="Infrastructure\ISortableProperties.cs" />
     <Compile Include="Infrastructure\ISortableProperty.cs" />
```
Aleksanderis: Is this still needed? Seems like you refactored all the places where you added that previously.

src/api/Shrooms.Contracts/Infrastructure/ISortableProperties.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Shrooms.Contracts.Infrastructure
+{
+    public interface ISortableProperties
```
Aleksanderis: Can it be named just `ISortable` (like existing `IPageable`)?
Currently it's implemented by DTO/ViewModels, which have a lot of properties - so interface name is confusing.

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
```diff
@@ -11,8 +10,74 @@ public static class DbSetExtensions
     {
         private const string DefaultSortDirection = "asc";
```
Aleksanderis: I'm starting to think that these extensions more likely belong to data access layer.
Another thing, that it looks like it extends `IQueryable` rather than `DbSet` (because of `this IQueryable<>` parameters everywhere).

jusrus01: Shrooms.Domain.Services project does not have a reference to Shrooms.DataLayer.DAL (only Shrooms.DataLayer.EntityModels which is another project). Would you like me to add reference to DAL and move the extensions? (Also I renamed this from DbSetExtensions to QueryableExtensions)

src/api/Shrooms.Contracts/Infrastructure/ISortableProperties.cs
jusrus01: Renamed.

src/api/Shrooms.Contracts/Shrooms.Contracts.csproj
jusrus01: Removed.

src/api/Shrooms.Domain/Extensions/DbSetExtensions.cs
Aleksanderis: Ok, let's leave it in Service layer then.

Aleksanderis (approved):

PR #219, opened 2022-07-13
Cleaned up code quality issues

Aleksanderis:


src/api/Shrooms.Domain/Services/FilterPresets/FilterPresetService.cs
```diff
@@ -153,11 +153,11 @@ private IQueryable<FilterDto> GetFiltersQueryByFilterType(FilterType filterType)
             };
         }
 
-        private async Task<IEnumerable<FilterPresetDto>> DeleteAsync(ManageFilterPresetDto manageFilterPresetDto)
+        private async Task<IList<FilterPresetDto>> DeleteAsync(ManageFilterPresetDto manageFilterPresetDto)
```
jusrus01, line 156: Could you explain why using IList<> is better than IEnumerable<>? I know that to get data with EF you would call .ToList(), but here I don't do that.

Aleksanderis: Because it's a private method and where it's used it's clear that it will have to do `ToList()` at least once, but maybe more.
Therefore every time it will have to enumerate. In such cases it's better to convert explicitly to list once.

In public methods - it's less evident and more often it's better to leave it to the user how it will be enumerated.

jusrus01 (approved):

PR #221, opened 2022-07-15
Improved event reports

jusrus01:
Added pagination, sorting and filtering by date to event report participant visited events.

### DIFF
--- a/src/api/Shrooms.Premium/Domain/Services/Events/List/EventListingService.cs
+++ b/src/api/Shrooms.Premium/Domain/Services/Events/List/EventListingService.cs
@@ -133,6 +133,31 @@ namespace Shrooms.Premium.Domain.Services.Events.List
 
             return events;
         }
+        
+        public async Task<IPagedList<EventVisitedReportDto>> GetEventParticipantVisitedReportEventsAsync(EventParticipantVisitedEventsListingArgsDto visitedArgsDto, UserAndOrganizationDto userOrg)
+        {
+            var eventTypesLength = visitedArgsDto.EventTypeIds.Count();
+            var visitedEvents = await _eventsDbSet
+                .Include(e => e.EventParticipants)
+                .Where(e => e.EventParticipants.Any(participant => participant.ApplicationUserId == visitedArgsDto.UserId) &&
+                                    e.EndDate < DateTime.UtcNow &&
+                                    e.OrganizationId == userOrg.OrganizationId &&
+                                    (eventTypesLength == 0 || visitedArgsDto.EventTypeIds.Contains(e.EventType.Id)) &&
+                                    e.StartDate >= visitedArgsDto.StartDate && 
+                                    e.EndDate <= visitedArgsDto.EndDate)
+                .Select(e => new EventVisitedReportDto
+                {
+                    Id = e.Id,
+                    Name = e.Name,
+                    TypeName = e.EventType.Name,
+                    StartDate = e.StartDate,
+                    EndDate = e.EndDate
+                })
+                .OrderByPropertyNames(visitedArgsDto)
+                .ToPagedListAsync(visitedArgsDto.Page, visitedArgsDto.PageSize);
+
+            return visitedEvents;
+        }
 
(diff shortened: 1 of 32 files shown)


src/api/Shrooms.Premium/Domain/Services/Events/List/EventListingService.cs
```diff
@@ -188,7 +210,12 @@ public async Task<IPagedList<EventParticipantReportDto>> GetReportParticipantsAs
                             StartDate = visited.StartDate,
                             EndDate = visited.EndDate
                         })
-                        .ToList()
+                        .ToList(),
+                    TotalVisitedEventCount = _eventsDbSet
+                        .Count(e => e.EventParticipants.Any(participant => participant.ApplicationUserId == p.ApplicationUser.Id) &&
+                                    e.EndDate < DateTime.UtcNow &&
+                                    e.OrganizationId == userOrg.OrganizationId &&
+                                    (eventTypesLength == 0 || reportArgsDto.EventTypeIds.Contains(e.EventType.Id)))
```
jusrus01, line 218: Tried to export this LINQ to Expression<Func<Event, bool>> (in this method we have the same expression twice) function, however I kept getting this error.
Screenshot 2022-07-14 104859

SauliusStankevicius: Which EF version does Simoona use? it seems there is a bug with making queries like that for navigation properties in EF 6, and also what happens if you only move the bool check inside the Any() to a separate method?

src/api/Shrooms.Premium/Presentation/WebViewModels/ValidationAttributes/DateTimeLessThanPresentDateAttribute.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.ComponentModel.DataAnnotations;
+
+namespace Shrooms.Premium.Presentation.WebViewModels.ValidationAttributes
+{
+    [AttributeUsage(AttributeTargets.Property)]
+    public class DateTimeLessThanPresentDateAttribute : ValidationAttribute
+    {
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            if (value == null || ((DateTime)value) <= DateTime.UtcNow)
```
SauliusStankevicius: Does UI send dates in UTC? and maybe better to validate if value is of type DateTime first, as this one can throw exception if attribute was set on some other property

src/api/Shrooms.Premium/Presentation/WebViewModels/Events/EventParticipantVisitedEventsListingArgsViewModel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        [Range(1, int.MaxValue)]
+        public int PageSize { get; set; } = EventsConstants.EventsDefaultPageSize;
+        
+        public string SortByProperties { get; set; }
+
+        public IEnumerable<int> EventTypeIds { get; set; }
+
+        [Required]
+        [DateTimeLessThanDateTime("EndDate")]
```
SauliusStankevicius: could `nameof` be used in here?

src/api/Shrooms.Premium/Presentation/WebViewModels/ValidationAttributes/DateTimeLessThanPresentDateAttribute.cs
jusrus01: It does send UTC. At least everything related with events and profile editing. Also, is that check necessary when the attribute name clearly states that it has to be used with DateTime?

SauliusStankevicius: the problem is that nobody is preventing it to be used on different property type

src/api/Shrooms.Premium/Domain/Services/Events/List/EventListingService.cs
jusrus01: Version is 6.1.3. 

Exporting Any() I get:
image

SauliusStankevicius (approved):

PR #225, opened 2022-08-01
Added email notification when new lottery is created

jusrus01:
Message looks like this:
image

Also can disable email notifications: 
image


jusrus01:
Now looks like this
image

SauliusStankevicius (approved):

PR #228, opened 2022-08-11
Added gifting lottery tickets to others

jusrus01:


src/api/Shrooms.Premium.Tests/DomainService/Validators/LotteryValidatorTests.cs
```diff
@@ -0,0 +1,393 @@
...
+
+namespace Shrooms.Premium.Tests.DomainService.Validators
+{
+    [TestFixture]
+    public class LotteryValidatorTests
+    {
+        private ISystemClock _systemClock;
+        private DbSet<ApplicationUser> _usersDbSet;
+
+        private LotteryValidator _sut;
```
SauliusStankevicius, line 25: the name `_sut` sounds strange

src/api/Shrooms.Premium/Domain/Services/Lotteries/LotteryAbortJob.cs
```diff
@@ -1,28 +1,28 @@
...
+using System.Linq;
+using System.Threading.Tasks;
 
 namespace Shrooms.Premium.Domain.Services.Lotteries
 {
     public class LotteryAbortJob : ILotteryAbortJob
     {
         private readonly ILotteryService _lotteryService;
-        private readonly IParticipantService _participantService;
+        private readonly ILotteryParticipantService _participantService;
```
SauliusStankevicius: `_participantService` can also be renamed to `_lotteryParticipantService`

src/api/Shrooms.Premium.Tests/DomainService/Validators/LotteryValidatorTests.cs
jusrus01: Means "System under test". I noticed some of the tests had that name, and it made more sense to me to call it that since it makes it easier to distinguish from other fields in the test class. Should I call it **_systemUnderTest** or just **_lotteryValidator** like I used to do?

SauliusStankevicius: I think we can leave as is

PR #229, opened 2022-08-12
Fixed 401 in local storage

jusrus01:
Front-end application receives images with the use of ngSrc directive and using it we are not able to send Bearer tokens that's why in addition to those tokens we also save .AspNet.Cookies cookie that allows us to view images without the use of Bearer token. The problem was that when logging in through internal login (not through a provider, e.g., Google) the expiration date of AspNet.Cookies was set to **Session**, so when we close the browser, we **lose** that cookie (however, the Bearer token is still saved in local storage, and since we don't log in again, we don't receive .AspNet.Cookies again).

What I did was set the cookie expiration date to the same date as the Bearer token expiration date. I also removed the "Remember me" checkbox because it does not work. Regardless of what is selected, it will always behave as if we have pressed "Remember me", because all it did was set the IsPersistence parameter (I also removed that parameter) for the <api>/Account/SignIn() endpoint that only influences cookies (and since we do authentication with the use of Bearer tokens, "Remember me" logic should be implemented differently).

### DIFF
--- a/src/api/Shrooms.Contracts/Infrastructure/IApplicationSettings.cs
+++ b/src/api/Shrooms.Contracts/Infrastructure/IApplicationSettings.cs
@@ -10,6 +10,8 @@ namespace Shrooms.Contracts.Infrastructure
 
         int DefaultOrganizationId { get; }
 
+        int AccessTokenLifeTimeInHours { get; }
+
         bool IsProductionBuild { get; }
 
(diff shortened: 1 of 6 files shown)


Aleksanderis (approved):

Aleksanderis:
Really good finding! 👍

PR #256, opened 2022-12-29
Fix new lottery email notification

jusrus01:


src/api/Shrooms.Premium/Domain/Services/Email/Lotteries/LotteryNotificationService.cs
```diff
@@ -43,7 +43,7 @@ public async Task NotifyUsersAboutStartedLotteryAsync(LotteryStartedEmailDto sta
         {
             var userEmailsGroupedByTimeZone = await _usersDbSet
                 .Include(user => user.NotificationsSettings)
-                .Where(user => user.NotificationsSettings.CreatedLotteryEmailNotifications)
```
SauliusStankevicius, line 46: If notification settings are `null` I think we should not send emails and treat that setting with default bool value - `false`

PR #259, opened 2023-01-30
Add one-time event reminders

jusrus01:


src/api/Shrooms.Premium/Presentation/WebViewModels/Events/CreateEventViewModel.cs
```diff
@@ -18,16 +19,17 @@ public class CreateEventViewModel
         public string ImageName { get; set; }
 
         [Required]
-        public DateTime StartDate { get; set; }
```
SauliusStankevicius, line 21: Why these made `nullable`, it seems a bit strange that they have `Required` attribute, but you still except `null` values

jusrus01: [Required] applied on struct or a value type does not work, because they have default values. For example, if I don't provide a value for **StartDate** in my POST request, when it hits the controller method it will have a value that will be set, regardless if it exists in the request or not and because of that ModelState.IsValid will return **true** instead of false. By making it nullable, it will set default value as null and ModelState.IsValid will return **false**.

PR #271, opened 2023-05-24
Change email receivers from "to" to "bcc"

Nikanoras:
Suggestion not to send all emails with receivers as `to` but as `bcc`. By this we can prevent unnecessary user information sharing.

### DIFF
--- a/src/api/Shrooms.Infrastructure/Email/MailingService.cs
+++ b/src/api/Shrooms.Infrastructure/Email/MailingService.cs
@@ -105,7 +105,7 @@ namespace Shrooms.Infrastructure.Email
             mailMessage.From = new MailAddress(sender);
             foreach (var receiver in email.Receivers)
             {
-                mailMessage.To.Add(receiver);
+                mailMessage.Bcc.Add(receiver);
             }
 
             if (email.Attachment != null)


src/api/Shrooms.Infrastructure/Email/MailingService.cs
```diff
@@ -105,7 +105,7 @@ private MailMessage BuildMessage(EmailDto email, bool skipDomainChange = false)
             mailMessage.From = new MailAddress(sender);
             foreach (var receiver in email.Receivers)
             {
-                mailMessage.To.Add(receiver);
+                mailMessage.Bcc.Add(receiver);
```
Aleksanderis, line 108: It's a known issue with unnecessary information sharing when we send emails from Simoona. However changing it to "BCC" instead - is not a solution. Emails having no "To" recipient and a list of emails in "BCC" is often a marker for spam filters, and might also impact account reputation in Sendgrid.

I don't fully remember, but seems like Sendgrid even wasn't allowing to set a list of emails in "BCC" field.

If you would like to solve it - then we would need to investigate what Sendgrid suggests (I'm sure it's a common topic).
Using Sendgrid API instead of SMTP could be also an option. However SMTP should be still supported as a fallback, because Sendgrid is not used in all the cases.

Aleksanderis (changes requested):

PR #275, opened 2023-06-27
Configurable monthly kudos send limit

xspeedasx:
Added KudosAvailableToSendPerMonth setting to Web.config - if specified it overrides the default 20.

### DIFF
--- a/src/api/Shrooms.Presentation.Api/Web.config
+++ b/src/api/Shrooms.Presentation.Api/Web.config
@@ -31,6 +31,7 @@
     <add key="CorsOrigin" value="*" />
     <add key="EmailEnabled" value="true" />
     <add key="DefaultOrganizationId" value="1" />
+    <add key="KudosAvailableToSendPerMonth" value="30" />
     <!--
       External Login Providers Properties
       needed for social logins
--- a/src/api/Shrooms.Presentation.Api/Web.config
+++ b/src/api/Shrooms.Presentation.Api/Web.config
@@ -31,7 +31,7 @@
     <add key="CorsOrigin" value="*" />
     <add key="EmailEnabled" value="true" />
     <add key="DefaultOrganizationId" value="1" />
-    <add key="KudosAvailableToSendPerMonth" value="30" />
+    <add key="KudosAvailableToSendPerMonth" value="20" />
     <!--
       External Login Providers Properties
       needed for social logins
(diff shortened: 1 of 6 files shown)


src/api/Shrooms.Presentation.Api/Web.config
```diff
@@ -1,342 +1,343 @@
...
+  <appSettings>
+    <add key="IsProductionBuild" value="false" />
+    <add key="webpages:Version" value="3.0.0.0" />
+    <add key="webpages:Enabled" value="false" />
+    <add key="PreserveLoginUrl" value="true" />
+    <add key="UnobtrusiveJavaScriptEnabled" value="true" />
+    <add key="CorsOrigin" value="*" />
+    <add key="EmailEnabled" value="true" />
+    <add key="DefaultOrganizationId" value="1" />
+    <add key="KudosAvailableToSendPerMonth" value="30" />
```
Aleksanderis: This is kind of a breaking change for currently deployed Simoona instances, since it will override an old default of 20 with next deployment if no action is taken. Either remove this setting from default web.config (and provide it in web.release.config used for your deployments), or change it to an old default value - 20.

Also in general - Simoona is multi-tenant solution (at least main parts of it). For example main Visma instance is hosting it for two organizations/tenants currently (in the past there were even more). Providing values in web.config would set settings for all tenants, which is not good.

Maybe monthly kudos amount is not so critical thing, but in general this pattern should be avoided. For now, some of such settings are stored in `Organizations` database table.

xspeedasx: I agree that this is just a relocation of the previously hard coded constant into configuration.
Ideally this should be configurable from UI for each tenant. And could even be configurable per-role. 
But as a temporary measure to enable single-tenant users to configure it globally it should be OK.

Speaking about the setting in question: 
I assume that usually when updating to a new version users don't blindly use a new `Web.config`, since it can reset all of their settings back to default.

Unless they update the source and rebuild in which case only using smart merge would add this new line to the config.

Maybe I should change this to be a commented out line to communicate better that this setting is optional.

Aleksanderis: > But as a temporary measure to enable single-tenant users to configure it globally it should be OK.

Such "temporary measures" become "permanent" right on time they are merged in. :)

> I assume that usually when updating to a new version users don't blindly use a new Web.config, since it can reset all of their settings back to default.

Everything what is pushed to remote repository accidentally or non-accidentally can be used blindly. 
I suggest just change the new value in web.config to a previous default value (which was hardcoded).
This won't introduce any unexpected behaviour for the ones who are not concerned about this change.

xspeedasx: OK, changed.

Aleksanderis (approved):

Issue #279, opened 2023-06-29
Remove Premium package


Nikanoras:
Hey @Aleksanderis ,
I wanted to ask if it would make sense to remove separate premium project and add it to all other "free-mium" stuff, as Simoona is fully open-source now? 
Why am I asking this, is because I started to investigate build issues with project it self. When you try to build all solution, you get lots of random errors from Shrooms.Premium project. My guess for the reason is the indirect project reference. As I checked Shrooms.Premium build output is being placed in Api project Extensions folder, and that's how premium is integrated. Please correct me if I'm wrong on this part. 

Kindly,
Nikanoras

Aleksanderis:
I would like to keep Premium features separate. Maybe the name is not correct anymore, but having these features separately kind of adds one stage of separation/modularization, and merging them back would be a step backwards.
However I think it can be changed so, that they are loaded not as extensions, but referenced as usual project reference.
It should solve project building issues.

Nikanoras:
Will do so.

Aleksanderis:
#284 was merged, so closing the issue.

PR #284, opened 2023-07-03
Add premium project reference

Nikanoras:
Hey,
As I mentioned in issue #279 I added a direct reference from Premium to Representation.Api project. To do that I created a Representation.Common project, which contains all shared components between these two projects. As I checked the app, it seems to work fine. Not sure about the Hangfire setup in `IocBootstrapper.cs` line:149-160. "Needed for Hangfire to process jobs from extension assemblies", now that we are using a direct reference, this code is no longer needed as I understand.


Aleksanderis:
> Hey, As I mentioned in issue #279 I added a direct reference from Premium to Representation.Api project. To do that I created a Representation.Common project, which contains all shared components between these two projects. As I checked the app, it seems to work fine. Not sure about the Hangfire setup in `IocBootstrapper.cs` line:149-160. "Needed for Hangfire to process jobs from extension assemblies", now that we are using a direct reference, this code is no longer needed as I understand.

It looks like a big change (not just because of amount files changed), which would need some better testing, including possibly needed changes for CI/CD pipeline. Unfortunately we don't have any capacity to do it now, so I cannot promise when we would be able to merge it.

Nikanoras:
I understand. Please take your time.

Aleksanderis (approved):

PR #285, opened 2023-07-03
Configurable email building strategies

xspeedasx:
Added setting for choosing a strategy for building email messages when multiple recipients are provided:
  - **AllTo** [default]: All recipients are put into "To" field.
  - **AllBcc**: All recipients are put into "Bcc" field and sender's email (noreply@simoona.com) is put into "To" field.
  - **SingleTo**: Single email per-recipient is sent with recipient being in "To" field.

Refactored `MailingService` for more testability,
Added tests for email building strategies


src/api/Shrooms.Presentation.Api/Web.config
```diff
@@ -30,8 +30,11 @@
     <add key="UnobtrusiveJavaScriptEnabled" value="true" />
     <add key="CorsOrigin" value="*" />
     <add key="EmailEnabled" value="true" />
+    <!-- Available strategies: AllTo, AllBcc, SingleTo -->
+    <add key="EmailBuildingStrategy" value="AllBcc" />
```
Aleksanderis: Same as previously - please leave current behaviour (AllTo) as default one.

src/api/Shrooms.Infrastructure/Email/ISmtpService.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// Sends mail messages via SMTP asynchronously.
+        /// </summary>
+        /// <param name="messages">Message collection for sending.</param>
+        /// <returns>A <see cref="Task"/> that represents asynchronous operation.</returns>
+        Task SendAsync(IEnumerable<MailMessage> messages);
+
+        /// <summary>
+        /// Determines if SMTP configuration is present and valid.
+        /// </summary>
+        bool HasSmtpServerConfigured();
```
Aleksanderis:
```suggestion
        bool IsSmtpServerConfigured();
```

I would say it would be more natural name. Otherwise it's not clear who "has". Even in comment you wrote "... configuration IS present". (I know this name was there before, but since you are refactoring it anyway.)

src/api/Shrooms.Infrastructure/Email/MailingService.cs
```diff
@@ -1,47 +1,28 @@
...
 namespace Shrooms.Infrastructure.Email
 {
     public class MailingService : IMailingService, IIdentityMessageService
     {
+        private readonly EmailBuildingStrategy _emailBuildingStrategy;
+        private readonly ISmtpService _smtpService;
         private readonly TelemetryClient _telemetryClient;
 
-        public MailingService()
+        public MailingService(ISmtpService smtpService, IApplicationSettings appSettings)
```
Aleksanderis: Injecting `ISmtpService` makes MailingService also Smtp, while it could accept some other abstraction, like some service which would send email using SendGrid API. Perhaps `ISmptService` could be something like `IMailSenderService` and `SmtpService` as one of its' implementation.
Yes, in this case there would be a bit too much abstractions over mail sending (IMailingService, IMailSenderService), but it would be at least more extensible. Also it could be an argument, that `IMailingService` is application-related service, while `IMailSenderService` is infrastructure-related service.

xspeedasx: That's a good point. I've forgot about the architectural plan during implementation.. One of the main reasons of abstracting this into a service is to be able to plug in other notification services like SendGrid or even some SMS service like Twilio :)
I'll rename the interface and it's methods to better suit the abstract role it is meant for.

src/api/Shrooms.Presentation.Api/Web.config
xspeedasx: This time it was unintentionally left as `AllBcc` 😄 changed it during testing and it slipped through the pre-commit overview...

Aleksanderis (approved):

PR #288, opened 2023-07-11
Make expandable menus

Nikanoras:
Hi,
Added a new feature to expand/collapse side menu. On initial load all menus are collapsed, except the one which has the active page on.
image

Here are all menus expanded:
image

All collapsed:
image


src/webapp/src/client/styles/style/_layout.less
```diff
@@ -70,14 +70,14 @@
...
-        a {
+        > a {
             animation: fadein 0.5s;
             border-left: 0px;
             color: @base-color;
             font-size: 13px;
             width: 100%;
             line-height: @line-height-base;
-            padding: 6px 20px;
+            padding: 6px 0px;
```
Aleksanderis:
```suggestion
            padding: 6px 0;
```

Aleksanderis (commented):
> On initial load all menus are collapsed

Collapsible menus are nice and cool, but as an user I wouldn't like such strict default behaviour. It's an open discussion which menus could be collapsed by default, but for example entries in Activities and Company are used quite often, so introducing an extra click to access them - wouldn't improve UX. 

Suggestion for further changes:
1. Leave default behavior as it is now - i.e. all menus are expanded by default.
2. Store expanded/collapsed state for each menu in a browser cookie, so that user would collapse what he wants and it would be persisted at least in that way.
3. Otherwise introduce some more flexible functionality how default expanded/collapsed states could be configured per menu.

Nikanoras:
Added persistence to expand/collapse so user can make menu setup as he desires. Also, the menus are automatically expanded when it contains an active link.

src/webapp/src/client/app/layout/left-menu/left-menu.controller.js
```diff
@@ -114,5 +130,9 @@
         function isSidebarOpen() {
             return leftMenuService.getStatus();
         }
+
+        function isMenuExpanded(groupName) {
```
Aleksanderis: Is that empty method really needed here?

src/webapp/src/client/app/common/expandable-menu/expandable-menu.component.js
```diff
@@ -0,0 +1,35 @@
...
+        });
+
+    function expandableMenuController() {
+        let vm = this;
+
+        vm.onExpandClick = onExpandClick;
+
+        const wasExpanded = localStorage.getItem(vm.menuName);
+        if (wasExpanded === 'true') {
+            vm.isExpanded = true;
```
Aleksanderis: So the default (when nothing stored in localstorage yet) menu will be collapsed (i.e. `wasExpanded` will be undefined/false)?

As mentioned previously, unless you don't want to introduce some configuration for that (I would say it doesn't worth the effort) - we would like to keep default state for menus as expanded.
I agree that perhaps "External links" (usually the biggest menu group) could be collapsed by default, but it would be good to keep "Activities" and "Companies" as for instance expanded by default.

src/webapp/src/client/app/common/expandable-menu/expandable-menu.component.js
```diff
@@ -0,0 +1,35 @@
...
+            controller: expandableMenuController,
+            controllerAs: 'vm'
+        });
+
+    function expandableMenuController() {
+        let vm = this;
+
+        vm.onExpandClick = onExpandClick;
+
+        const wasExpanded = localStorage.getItem(vm.menuName);
```
Aleksanderis: It would be good to add some prefix for the localstorage item to make it more clear when you would look at/debug localstorage. (And as general practice - menu names are quite generic, so names could conflict with other entries.)

Aleksanderis (changes requested):

src/webapp/src/client/app/layout/left-menu/left-menu.controller.js
Nikanoras: Removed.

src/webapp/src/client/app/common/expandable-menu/expandable-menu.component.js
Nikanoras: Reversed the logic to expand everything by default.

src/webapp/src/client/app/common/expandable-menu/expandable-menu.component.js
Nikanoras: Added a postfix. If needed can change it to prefix.

src/webapp/src/client/app/common/collapsable-menu/collapsable-menu.component.js
```diff
@@ -0,0 +1,35 @@
...
+                isCollapsed: '<'
+            },
+            templateUrl: 'app/common/collapsable-menu/collapsable-menu.html',
+            controller: collapsableMenuController,
+            controllerAs: 'vm'
+        });
+
+    function collapsableMenuController() {
+        const vm = this;
+        const localStorageName = vm.menuName + 'isCollapsed';
```
Aleksanderis: For possible conflicting names issues it's good enough, but prefix (something like `Menu`) would add also add more clarity for next generations. :)

Nikanoras: Added prefix.

Aleksanderis (approved):

PR #290, opened 2023-07-19
Migrate packages.config

Nikanoras:
Hi,
Started the migration from packages.config to PackageReference. This is one of prerequisite steps to migrate this app to .NET Core someday. As I tested the app works fine. At the moment I could not fully migrate from packages.config, because of indirect reference from Premium package. I tested this branch on #284, and it works. So this PR needs to wait until that one is merged.

Btw, I would suggest to use squash for PR merging, as it will be a single commit and it will be easier to revert if a need rises.


Aleksanderis:
We already were working on .NET 6 migration. It's also on hold now, but there was a lot of work done already, so I would suggest to start from there.
It's in the fork of one of our developers - https://github.com/jusrus01/simoona/tree/dotnet-6-migration-basic-projects

PR #306, opened 2026-07-08
Add emoji reactions to kudos logs

edgarasLegusVisma:
## What

Lets users react to entries in the recent-kudos wall widget with the same emoji reactions that posts and comments already support.

- `KudosLog` gets a `Likes` column — the same JSON-serialized `LikesCollection` owned-type mapping used by `Post` and `Comment`.
- New endpoint: `PUT /Kudos/Like` with `{ id, type }` (mirrors `PUT /Post/Like`), gated by `BasicPermissions.Kudos`. Toggles per `(user, type)`; only approved logs in the caller's organization can be reacted to.
- `WallKudosLogViewModel` (returned by `/Kudos/GetLastKudosLogRecords` and inside `/WallWidgets/Get`) now carries `id` and `likes` (`{ userId, fullName, pictureId, type }`, deleted reactors filtered out) so clients can render and toggle reactions.

## Migration note

`AddKudosLogLikes` adds only the nullable `KudosLogs.Likes` column. The regenerated model snapshot additionally records the `IsDeleted` columns on `BadgeTypes`/`BadgeLogs`/`BadgeCategories` that the previous snapshot was missing — those columns already exist in the database (created by `InitialBaseline`), so no schema operations are emitted for them. This brings the snapshot back in sync so future migrations stop trying to re-add them.

## Consumer

The Simoona 2.0 frontend already ships the (dormant) reactions UI on the recent-kudos widget — it activates automatically once this response contract is live. Frontend contract reference: `KudosWallLogEntry` in `src/types/wall-widgets.ts`.

## Tests

7 new `KudosServiceTests` covering toggle add/remove, multi-type reactions, unknown/pending/cross-org logs, and like enrichment (incl. deleted-reactor filtering). Full suite: 556/556 passing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/api/Shrooms.Presentation.Common/Controllers/Kudos/KudosController.cs
```diff
@@ -385,6 +387,29 @@ public async Task<IEnumerable<WallKudosLogViewModel>> GetLastKudosLogRecords()
...
 
+        [HttpPut]
+        [PermissionAuthorize(Permission = BasicPermissions.Kudos)]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        public async Task<IActionResult> Like(AddLikeViewModel addLikeViewModel)
+        {
+            if (!ModelState.IsValid)
+            {
+                return BadRequest();
+            }
```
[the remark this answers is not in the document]
arturasvisma: Agree with this

src/api/Shrooms.Domain/Services/Kudos/KudosService.cs
```diff
@@ -293,12 +295,90 @@ public async Task<IEnumerable<WallKudosLogDto>> GetLastKudosLogsForWallAsync(Use
                     log.KudosSystemType != KudosTypeEnum.Minus &&
```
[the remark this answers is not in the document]
arturasvisma: Agree with this as well

src/api/Shrooms.Domain/Services/Kudos/KudosService.cs
```diff
@@ -293,12 +295,90 @@ public async Task<IEnumerable<WallKudosLogDto>> GetLastKudosLogsForWallAsync(Use
...
                     log.KudosSystemType != KudosTypeEnum.Refund &&
                     log.OrganizationId == userAndOrg.OrganizationId)
-                .Join(_usersDbSet, l => l.CreatedBy, s => s.Id, MapKudosLogToWallKudosLogDto())
-                .OrderByDescending(log => log.Created)
+                .Join(_usersDbSet, l => l.CreatedBy, s => s.Id, (log, sender) => new { Log = log, Sender = sender })
```
arturasvisma: 🔴 **`Include(Employee)` may be dropped across this `.Join(...)`.**

The projection into an anonymous type here means `x.Log.Employee` is read **in memory** further down (in the `Receiver` mapping). `Include` composed with a shape-changing `.Join(...)` is a known EF Core "ignored includes" case, and `ChangeTracker.LazyLoadingEnabled = false`, so there is no lazy-load fallback. If the include is not carried through the join, `Receiver.FullName`/`Receiver.Id` become `null` for **every** wall entry.

The old code translated the whole projection to SQL, so Employee always flowed through — this is a behavior change.

⚠️ The `Should_Return_Wall_Kudos_Logs_With_Id_And_Likes` test cannot catch this: the fake DbSet returns the in-memory object graph with `Employee` pre-assigned, bypassing EF materialization. **Please verify against a real database** (or restore a SQL-side projection that selects the employee fields).

src/api/Shrooms.Domain/Services/Kudos/KudosService.cs
```diff
@@ -293,12 +295,90 @@ public async Task<IEnumerable<WallKudosLogDto>> GetLastKudosLogsForWallAsync(Use
...
+                    {
+                        FullName = x.Sender.FirstName + " " + x.Sender.LastName,
+                        Id = x.Log.KudosSystemType == KudosTypeEnum.Send ? x.Log.CreatedBy : null
+                    },
+                    Likes = MapLikesToDto(x.Log.Likes, likers)
+                })
+                .ToList();
+        }
+
+        public async Task ToggleLikeAsync(AddLikeDto addLikeDto, UserAndOrganizationDto userOrg)
```
arturasvisma, line 338: 🟡 **Lost-update race.** Two concurrent like/unlike requests on the same log both read the log, both mutate the JSON `Likes` collection, and the second `SaveChangesAsync` overwrites the first → one reaction silently lost. There is no concurrency token on the `Likes` column and no serialization here.

For comparison, `PostService.ToggleLikeAsync` wraps its body in a `SemaphoreSlim`. Worth adding similar protection (a user double-clicking, or two users reacting at once, is realistic).

src/api/Shrooms.Domain/Services/Kudos/KudosService.cs
```diff
@@ -293,12 +295,90 @@ public async Task<IEnumerable<WallKudosLogDto>> GetLastKudosLogsForWallAsync(Use
...
+
+        private static IEnumerable<LikeDto> MapLikesToDto(LikesCollection likes, IEnumerable<ApplicationUser> users)
+        {
+            if (likes == null)
+            {
+                return new List<LikeDto>();
+            }
+
+            return likes
+                .Select(like => new { User = users.FirstOrDefault(user => user.Id == like.UserId), Like = like })
```
arturasvisma: 🟢 **O(likes × users) lookup.** `users.FirstOrDefault(user => user.Id == like.UserId)` runs a linear scan of the `IEnumerable` per like. Minor, but building a `Dictionary<string, ApplicationUser>` keyed by `Id` once avoids the repeated scan.

src/api/Shrooms.Presentation.Common/Controllers/Kudos/KudosController.cs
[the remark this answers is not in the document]
edgarasLegusVisma: Applied

src/api/Shrooms.Domain/Services/Kudos/KudosService.cs
edgarasLegusVisma: Added the usage of SemaphoreSlim

src/api/Shrooms.Domain/Services/Kudos/KudosService.cs
edgarasLegusVisma: EF Core 10 actually did carry the Include even before the fix (verified), but the explicit projection makes it guaranteed by construction.

src/api/Shrooms.Domain/Services/Kudos/KudosService.cs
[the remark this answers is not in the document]
edgarasLegusVisma: Should be OK now

arturasvisma (approved):

PR #307, opened 2026-07-17
Feat/custom emoji registry

KristupasLunskas:
This pull request introduces support for custom emojis in the system. It adds the necessary data models, database migration, validation logic, and permission constants to enable organizations to manage their own custom emoji sets. The changes include new DTOs and entity models, database schema updates, validation rules, and permission assignments.

**Custom Emoji Data Model and Database Integration:**

* Added the `CustomEmoji` entity model and its configuration, including properties for name, blob storage reference, author, and organization, with uniqueness enforced per organization and soft deletion support (`CustomEmoji.cs`, `CustomEmojiConfig.cs`, `ShroomsDbContext.cs`, migration and snapshot updates). [[1]](diffhunk://#diff-23825cada60955830d1f87a02663b40446384630aea142208af02cf886ea525eR1-R19) [[2]](diffhunk://#diff-b4c9d3b13ad07ebaa91b1429cfc0ee10e0e1fa389ce74e3a372f87b8db504492R1-R39) [[3]](diffhunk://#diff-cf90a99ac46e833ff1910be8c28555a4478d6d8c92e2d5884cf230b641d28e31R18) [[4]](diffhunk://#diff-cf90a99ac46e833ff1910be8c28555a4478d6d8c92e2d5884cf230b641d28e31R161-R162) [[5]](diffhunk://#diff-cf90a99ac46e833ff1910be8c28555a4478d6d8c92e2d5884cf230b641d28e31R248) [[6]](diffhunk://#diff-b408126783193459b488b064887399f7f29d641030f786994a1c164d7e44cf6dR1-R127) [[7]](diffhunk://#diff-482c3e80596a09f820c58afa7edf950cebc7f14b998006c3a0c04cb8bc06567eR1129-R1180) [[8]](diffhunk://#diff-482c3e80596a09f820c58afa7edf950cebc7f14b998006c3a0c04cb8bc06567eR3841-R3859)

* Added new DTOs for transferring custom emoji data: `CustomEmojiDto`, `CustomEmojiListDto`, and `NewCustomEmojiDto`. [[1]](diffhunk://#diff-7c37ffdbe7a34900ed78bbb44cf2a5d42287507470c18a0dcc5a7d119ed30c1bR1-R13) [[2]](diffhunk://#diff-d6996ac612986d049b79f1546a9485f057037f4f0694b00946fc0cbe7143329fR1-R11) [[3]](diffhunk://#diff-0384a409c03d7b7c541ba649a1813a5b380a87a172f47cc2b647606774daa1fdR1-R15)

**Permissions and Constants:**

* Introduced new permission constants for custom emoji management and assigned them to relevant roles during migration (`AdministrationPermissions`, `BasicPermissions`, migration SQL). [[1]](diffhunk://#diff-b780603b99525aca6870c08735ff861911c19243931e9b61010a9803dae97404R31) [[2]](diffhunk://#diff-418c70b429b3b94a2d530930f5048b92eecfc112fe0cb995b26b12cca28539d3R42) [[3]](diffhunk://#diff-b408126783193459b488b064887399f7f29d641030f786994a1c164d7e44cf6dR1-R127)
* Added error codes, validation constants (name length, regex), and API constants (max file size, dimensions, allowed content types) specific to custom emojis. [[1]](diffhunk://#diff-2aac7bd215650b5879f028cd1dc7f6ad876b6d82268bf8c39f14931009a61602R64-R68) [[2]](diffhunk://#diff-ff04795834e0a3d36ab5192c2ea57b484dd5059b74bc6c20227d9029ff348c43R23-R25) [[3]](diffhunk://#diff-6934c342fc9199ea4f4d26c8451d4586ec0194b64efdb7e674ea4209e1b4c734R28-R30)

**Validation Logic:**

* Implemented `ICustomEmojiValidator` and its concrete class, providing validation for emoji name format, uniqueness within an organization, and image file type/size/dimensions using ImageSharp. [[1]](diffhunk://#diff-d7a1dbf22a01bd6c5b988ae98fc050f4704e7ff6053fbb053245471c5f19b638R1-R14) [[2]](diffhunk://#diff-4db2352c60dcbd04a93aa6cfc058cc6faf440ce417c3a8f0dc95cc1613b5609aR1-R72) [[3]](diffhunk://#diff-04d3161cfd29c7ed12583eee7e4d94e8dc0e14a333441e5270438774a259028dR13)

These changes lay the groundwork for custom emoji functionality, ensuring data integrity, security, and proper validation throughout the system.


src/api/Shrooms.DataLayer.EntityModels/Models/Emoji/CustomEmoji.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        [Required]
+        [StringLength(50)]
+        public string Name { get; set; }
+
+        [Required]
+        public string BlobName { get; set; }
+
+        [Required]
+        public string AuthorId { get; set; }
```
KristupasLunskas: **Question: is `AuthorId` redundant with `CreatedBy`?**

Kept deliberately, but open to discussion:

- `CreatedBy` is untyped audit metadata: no FK, no index, and some save paths leave it null. Basing the **delete-own authorization check** on it felt fragile.
- `Post`/`Comment` set the precedent with their own explicit `AuthorId`.
- The `Author` navigation lets the future management page show the uploader without a string join.

Happy to drop it if we agree `CreatedBy` is reliable enough for an authorization decision.

src/api/Shrooms.DataLayer/DAL/EntityTypeConfigurations/CustomEmojiConfig.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            builder.HasIndex(x => new { x.OrganizationId, x.Name })
+                .IsUnique()
+                .HasFilter("[IsDeleted] = 0")
+                .HasDatabaseName("IX_CustomEmojis_OrganizationId_Name");
+
+            builder.HasOne(e => e.Organization)
+                .WithMany()
+                .HasForeignKey(e => e.OrganizationId)
+                .OnDelete(DeleteBehavior.Restrict);
```
KristupasLunskas, line 26: **Note: `DeleteBehavior.Restrict` on both FKs is deliberate** 

- A hard cascade from a user/org delete would physically destroy emoji rows the soft-delete design promises to keep (historical references for a future reactions consumer).
- Another `ON DELETE CASCADE` path onto `AspNetUsers`/`Organizations` risks SQL Server's *multiple cascade paths* migration failure.
- Matches existing convention (`ExternalLinkConfig` does the same).

No impact on normal flows: users and orgs are soft-deleted, so no physical DELETE ever hits these FKs. The constraint only fires on out-of-band hard deletes, where failing loudly is the intended behavior.

src/api/Shrooms.DataLayer/EFCoreMigrations/20260716093729_AddCustomEmojis.cs
```diff
@@ -0,0 +1,127 @@
...
+                column: "AuthorId");
+
+            migrationBuilder.CreateIndex(
+                name: "IX_CustomEmojis_OrganizationId_Name",
+                table: "CustomEmojis",
+                columns: new[] { "OrganizationId", "Name" },
+                unique: true,
+                filter: "[IsDeleted] = 0");
+
+            migrationBuilder.Sql(@"
```
KristupasLunskas, line 45: **Note: the raw SQL block is a deliberate permission seed**, not leftover test data.

- Inserts `CUSTOMEMOJI_BASIC` (scope `basic`) and `CUSTOMEMOJI_ADMINISTRATION` (scope `admin`), guarded by `IF NOT EXISTS`.
- Grants them by copying role assignments from `PICTURE_BASIC` / `WALL_ADMINISTRATION` holders, deduped with `NOT EXISTS` - the whole block is idempotent.
- `Down` mirrors it (RolePermissions -> Permissions -> drop table).

This is the codebase convention for permissions (`AddEventUsersPermission` is the exact precedent). Without the seed, no role holds the permission and every upload/delete returns 403.

src/api/Shrooms.Domain/Services/Emoji/CustomEmojiService.cs
```diff
@@ -0,0 +1,158 @@
...
+            _emojiListCache = emojiListCache;
+            _generationCache = generationCache;
+            _customEmojisDbSet = uow.GetDbSet<CustomEmojiEntity>();
+        }
+
+        public async Task<CustomEmojiListDto> GetAllAsync(UserAndOrganizationDto userOrg, string tenantName)
+        {
+            var generation = GetGeneration(userOrg.OrganizationId);
+
+            if (_emojiListCache.TryGetValue(userOrg.OrganizationId, out var entry) && entry.Generation == generation)
```
KristupasLunskas, line 48: **Note on cache semantics (deliberate):**

- The list cache is **per-node with no TTL**, inheriting the same single-instance assumption as the existing `PermissionService` cache. If the API is ever scaled out, both need a TTL safety net or distributed invalidation - conscious inheritance, not an oversight.
- The generation counter (`GetGeneration` / `BumpGeneration`) guards the single-node race: a GET that queried *before* a concurrent create/delete committed could otherwise cache a stale list *after* the invalidation, pinning stale data (and its ETag) until the org's next mutation. Fetched lists are cached only if the generation is unchanged since before the query; cache hits are honored only while their recorded generation matches.

src/api/Shrooms.Domain/Services/Emoji/CustomEmojiService.cs
```diff
@@ -0,0 +1,158 @@
...
+            var emoji = new CustomEmojiEntity
+            {
+                Name = emojiDto.Name,
+                BlobName = blobName,
+                AuthorId = userOrg.UserId,
+                OrganizationId = userOrg.OrganizationId
+            };
+
+            _customEmojisDbSet.Add(emoji);
+            await _uow.SaveChangesAsync(userOrg.UserId);
```
KristupasLunskas, line 92: **Accepted limitation: concurrent duplicate-name creates.**

Two simultaneous creates with the same name both pass the uniqueness check; the loser hits the filtered unique index and surfaces as a **500** (unmapped `DbUpdateException`) with one orphaned blob in storage. Data integrity is guaranteed by the index.

If this ever matters in practice: catch `DbUpdateException`, detect SQL errors 2601/2627, rethrow as `ValidationException(DuplicatesIntolerable)` -> 409.

src/api/Shrooms.Domain/Services/Emoji/CustomEmojiService.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            return MapToDto(emoji, tenantName.ToLowerInvariant());
+        }
+
+        public async Task DeleteAsync(int id, UserAndOrganizationDto userOrg)
+        {
+            var emoji = await _customEmojisDbSet
+                .FirstOrDefaultAsync(x => x.Id == id && x.OrganizationId == userOrg.OrganizationId && !x.IsDeleted);
+
+            if (emoji == null)
```
KristupasLunskas, line 104: **Note: the existence check is deliberately inline**, not a `CheckIfEmojiExists` validator method (despite the `BlacklistValidator` precedent).

A null check on a value the service itself just fetched encapsulates no rule - only the service knows what "absent" means for its own query. Routing it through the injected validator also made the not-found test tautological (stub the mock to throw, assert it threw). The validator keeps only rules with real content: name format, uniqueness, image checks.

src/api/Shrooms.Contracts.DataTransferObjects/Models/Emoji/CustomEmojiDto.cs
```diff
@@ -0,0 +1,13 @@
...
+{
+    public class CustomEmojiDto
+    {
+        public int Id { get; set; }
+
+        public string Name { get; set; }
+
+        public string Url { get; set; }
+
+        public string AuthorId { get; set; }
```
arturasvisma: CreatedBy?

src/api/Shrooms.Contracts.DataTransferObjects/Models/Emoji/CustomEmojiListDto.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace Shrooms.Contracts.DataTransferObjects.Models.Emoji
+{
+    public class CustomEmojiListDto
+    {
+        public IEnumerable<CustomEmojiDto> Emojis { get; set; }
+
+        public string ETag { get; set; }
```
arturasvisma, line 9: Maybe rename to `Tag`?

src/api/Shrooms.Domain/Services/Emoji/CustomEmojiService.cs
```diff
@@ -0,0 +1,158 @@
...
+                throw new ValidationException(ErrorCodes.ContentDoesNotExist, "Emoji does not exist");
+            }
+
+            if (emoji.AuthorId != userOrg.UserId &&
+                !await _permissionService.UserHasPermissionAsync(userOrg, AdministrationPermissions.CustomEmoji))
+            {
+                throw new UnauthorizedException();
+            }
+
+            _customEmojisDbSet.Remove(emoji);
```
arturasvisma, line 117: Delete only marks the row deleted, but the image file stays in storage forever - call _pictureService.RemoveImageAsync too.

src/api/Shrooms.Domain/Services/Emoji/CustomEmojiService.cs
```diff
@@ -0,0 +1,158 @@
...
+            return _generationCache.TryGetValue(organizationId, out var generation) ? generation : 0;
+        }
+
+        private void BumpGeneration(int organizationId)
+        {
+            while (true)
+            {
+                if (_generationCache.TryGetValue(organizationId, out var current))
+                {
+                    _generationCache.TryRemoveEntry(organizationId);
```
arturasvisma: Remove-then-add is not atomic and can drop the counter back to a smaller number - use AddOrUpdate instead.

src/api/Shrooms.Presentation.Api/Controllers/EmojiController.cs
```diff
@@ -0,0 +1,121 @@
...
+            var etag = $"\"{result.ETag}\"";
+            Response.Headers.ETag = etag;
+            Response.Headers.CacheControl = "private";
+
+            if (Request.Headers.IfNoneMatch.Contains(etag))
+            {
+                return StatusCode(304);
+            }
+
+            return Ok(_mapper.Map<IEnumerable<CustomEmojiDto>, IEnumerable<CustomEmojiViewModel>>(result.Emojis));
```
arturasvisma, line 48: Anyone signed in can list all emojis and read author ids. Not sure it is good idea to see for everyone?

src/api/Shrooms.Contracts.DataTransferObjects/Models/Emoji/CustomEmojiDto.cs
KristupasLunskas: Removed AuthorId completely. Replaced with CreatedBy.

src/api/Shrooms.Domain/Services/Emoji/CustomEmojiService.cs
KristupasLunskas: Changed to hard deletion.

src/api/Shrooms.Domain/Services/Emoji/CustomEmojiService.cs
KristupasLunskas: Added AddOrUpdate.

arturasvisma (approved):

arturasvisma (approved):

PR #322, opened 2026-08-21
fix(emails): drop the inline mark and link home from configuration

IgnasJank:
Gmail strips inline SVG and drops `animation`, so the header's SimoonaMark rendered as an empty bordered box for most recipients and stood still for the rest. A hosted raster would be the only remote fetch in the whole set and needs a publicly reachable URL, so the mark and its ten keyframe blocks are gone. The S/moona wordmark beside it is plain text and carries the brand everywhere.

The wordmark's link and the footer's www.simoona.com link were both hardcoded. The footer link is dropped, and the wordmark now takes ClientUrl - filled in by MailTemplate rather than by every caller that builds a view model - so it points at the client the sending environment is configured against instead of a fixed host. Golden renders bypass MailTemplate, so EmailTemplateSeeds carries the value for them, and the layout is left with no hardcoded URLs at all.


arturasvisma:
CLAUDE

HtmlToPlainTextConverter.cs:56 — HtmlDecode runs after tag stripping, so HTML-escaped markup reappears as literal tags in the text/plain part. Reproduced against the repo's own baseline: Lotteries.StartedLottery converts to <p>Two tickets to the summer festival.</p>.
MailingService.cs:127 — the converter is applied to every EmailDto.Body, including non-template bodies; SupportService's raw user message loses all its line structure and any <…> run in the text/plain part.
MailingService.cs:129 — a null Body now throws ArgumentNullException (verified) and escapes the catch (SmtpException) guard, where mailMessage.Body = null used to be accepted. Line 128 null-guards; line 129 doesn't.
EmailEmTagHelper.cs:16 — setting the style via output.Attributes.SetAttribute entity-encodes it, so the accent renders as font-family:&#x27;Instrument Serif&#x27;,… — the only style attribute in the set that isn't emitted literally.
EmailTemplateSeeds.cs:131 — the StartedLottery seed feeds <p>…</p> into a Razor-encoded slot, so the approved golden and the preview page show raw tags; the app actually stores that field as plain textarea text. Pre-existing, but it's the baseline this PR's new test reads.

arturasvisma (approved):

PR #325, opened 2026-08-24
Kudos admin - allow export by filtered type

edgarasLegusVisma:
When admin will filter the log by type - the excel export will honour it


arturasvisma:
@edgarasLegusVisma I guess UI should be more clear so that the export is related with filtering type.

arturasvisma (approved):

IgnasJank (approved):

PR #329, opened 2026-08-31
fix(kudos): make the kudos admin log sortable by sender

edgarasLegusVisma:


src/api/Shrooms.Domain/Services/Kudos/KudosService.cs
```diff
@@ -190,7 +190,7 @@ public async Task<KudosLogsEntriesDto<MainKudosLogDto>> GetKudosLogsAsync(KudosL
                 .Where(KudosServiceHelper.StatusFilter(options.Status))
                 .Where(KudosServiceHelper.UserFilter(options.SearchUserId))
                 .Where(KudosServiceHelper.TypeFilter(options.FilteringType))
-                .GroupJoin(_usersDbSet, log => log.CreatedBy, u => u.Id, KudosServiceHelper.MapKudosLogsToDto());
+                .SelectMany(log => _usersDbSet.Where(user => user.Id == log.CreatedBy).DefaultIfEmpty(), KudosServiceHelper.MapKudosLogsToDto());
```
arturasvisma, line 193: Sorting uses only one column, so rows with the same value can come back in a different order every time, and then paging can show one row twice and hide another one.

Could we add `Id` as a second sort key here so the order is always the same?

edgarasLegusVisma: Fixed, thanks!

PR #334, opened 2026-09-01
fix(events): let an event with no cover image be updated

SamantaValunte:
`ImageName` was `[Required]` on update but optional on create, so an event with
no cover image failed validation on every `PUT /Events/Update` and could not be
edited at all — hit while adding sign-up questions to one. Split out of #332.

A regression, not new permissiveness: until d327b1b2 (Feb 2023) update was
`UpdateEventViewModel : CreateEventViewModel` and inherited an optional
`ImageName`; decoupling them added `[Required]` to update only. Create, the
nullable column and the `wall-default.png` fallback always allowed no image.


src/api/Shrooms.Premium/Presentation/WebViewModels/Events/UpdateEventViewModel.cs
```diff
@@ -22,7 +22,10 @@ public class UpdateEventViewModel
         [StringLength(EventsConstants.EventNameMaxLength)]
         public string Name { get; set; }
 
-        [Required]
+        /// <summary>
```
arturasvisma, line 25: no comments is needed

arturasvisma (approved):

IgnasJank (approved):

PR #335, opened 2026-09-01
feat(Polls): implemented backend of polls functionality

arturasvisma:


IgnasJank:
Claude findings

1. Voters are never added to the poll's wall — PollService.cs:389
AddMemberToWallsAsync only stages _wallUsersDbSet.Add(...); it does not save. VoteAsync's only SaveChangesAsync runs before the call, so the membership rows are dropped when the context is disposed. Every other caller saves afterwards (ProjectsService.cs:307, WallService.AddModeratorAsync:463). Net effect: no voter ever joins the discussion wall.

2. Fixing #1 breaks anonymity — wall membership is a public list (GetWallMembersIdsAsync, wall UI). Joining voters on vote publishes exactly who voted. The rest of the design is careful here (ApplicationUserId = null, no timestamps on PollAnswer, shuffled insert order) — this one path undoes it. Skip the join for anonymous polls, or seed the wall with the whole audience up front.

3. Orphaned wall on failure — CreateAsync. CreateNewWallAsync commits its own save (WallService.cs:112); if the poll insert then throws, the wall is already persisted with nothing to clean it up. DeleteAsync has the mirror problem: the poll soft-delete commits, then DeleteWallAsync can throw and leave a deleted poll with a live wall. Neither pair is in a transaction.

4. UpdateAsync silently discards edits — when hasVotes, question/option/IsAnonymous changes are dropped and the caller still gets 200. The frontend gates this with ballotLocked, but a stale page or a direct call loses data with no signal. Throw instead.

5. catch (DbUpdateException) reports every DB failure as "You have already voted." — VoteAsync. Deadlocks, timeouts and FK errors all surface as that message. Nviolation (2601/2627) and rethrow.

6. GuardedAsync returns raw exception text as 400 — PollController.cs:150. Any EF InvalidOperationException becomes a 400 whose body is an internal message shown ." is a 400 rather than 404. That forcessrc/lib/api/polls.ts:11 to treat 400 as not-found, so genuine validation errors on Get silently become notFound().

7. Zero tests. The same branch ships ~1300 ly and none for polls, though PollServiceholds all the non-obvious rules: state machine, the canSeeResults matrix, anonymity, vote validation, the ballot lock.

Lower priority: PublishAsync silently rewritDays(1) without telling the admin; neitherlist endpoint pages (the whole org's polls are materialised on every wall render, since the widget calls /Poll/List); a title edit never propagates to the wall's Name/Description it seeded; UpdateAsync validates content and deadline before the authorization check. Also worth cmission seeding is intended — it copiesEVENT_* role grants, so a tenant with no event grants gets no poll access at all.

src/api/Shrooms.Presentation.Api/Controllers/PollController.cs
```diff
@@ -0,0 +1,172 @@
...
+            return await GuardedAsync(async () =>
+            {
+                await _pollService.CloseAsync(id, GetUserAndOrganization());
+                return true;
+            });
+        }
+
+        [HttpDelete]
+        [PermissionAuthorize(AdministrationPermissions.Poll)]
+        public async Task<IActionResult> Delete(int id)
```
edgarasLegusVisma, line 122: Seems like deleting a poll fails every time?...

Image

⨯ Error: Failed to delete the poll.
    at send (src\lib\actions\polls.ts:60:11)
    at async deletePollAction (src\lib\actions\polls.ts:125:3)
  58 |
  59 |   if (!res.ok) {
> 60 |     throw new Error(await readError(res, fallback));
     |           ^
  61 |   }
  62 |
  63 |   revalidatePolls(); {
  digest: '22520478'
}
 POST /en/administration/polls 500 in 383ms (next.js: 17ms, proxy.ts: 3ms, application-code: 363ms)
  └─ ƒ deletePollAction(2) in 39ms src/lib/actions/polls.ts

src/api/Shrooms.Domain/Services/Polls/PollService.cs
```diff
@@ -0,0 +1,705 @@
...
+            var state = !canManage || dto.Suggest
+                ? PollState.Pending
+                : dto.Publish ? PollState.Published : PollState.Draft;
+
+            if (state != PollState.Draft && dto.Deadline <= DateTime.UtcNow)
+            {
+                throw new ArgumentException("The deadline must be in the future.");
+            }
+
+            var wallId = await _wallService.CreateNewWallAsync(new CreateWallDto
```
edgarasLegusVisma, line 129: Codex - CreateAsync can leave an orphaned wall. The wall is created and committed before the poll is inserted, with no transaction around the pair, so any failure on the poll insert leaves a stray private wall.

src/api/Shrooms.Domain/Services/Polls/PollService.cs
```diff
@@ -0,0 +1,705 @@
...
+                Questions = BuildQuestions(dto.Questions, dto.UserId, now)
+            };
+
+            _pollDbSet.Add(poll);
+            await _uow.SaveChangesAsync(false);
+
+            return await GetPollAsync(poll.Id, dto, canManage);
+        }
+
+        public async Task UpdateAsync(UpdatePollDto dto, bool canManage)
```
edgarasLegusVisma, line 165: Codex - UpdateAsync silently discards edits once a poll has votes
PollService.cs:203 — when hasVotes, the submitted questions, options and IsAnonymous are dropped and the call still returns 200. The frontend mirrors this with ballotLocked so the UI hides it, but any drift between the two (or a direct API call) means a silent partial save. Given Update throws for other invalid states, rejecting here would be consistent.

src/api/Shrooms.Domain/Services/Polls/PollService.cs
```diff
@@ -0,0 +1,705 @@
...
+                {
+                    question.PollId = poll.Id;
+                    _questionDbSet.Add(question);
+                }
+            }
+
+            await _uow.SaveChangesAsync(false);
+        }
+
+        public async Task PublishAsync(PollReviewArgsDto args)
```
edgarasLegusVisma, line 249: Codex - PublishAsync silently rewrites a past deadline
PollService.cs:245 — if (poll.Deadline <= now) poll.Deadline = now.AddDays(1);. CreateAsync rejects a past deadline outright, so the same input is an error on create and a silent 24-hour extension on publish. An admin approving a suggestion that sat too long gets an arbitrary deadline with no indication.

src/api/Shrooms.Domain/Services/Polls/PollService.cs
```diff
@@ -0,0 +1,705 @@
...
+            var isOwnSuggestion = isAuthor &&
+                                  (poll.State == PollState.Pending || poll.State == PollState.Rejected);
+
+            if (!canManage && !isOwnSuggestion)
+            {
+                throw new InvalidOperationException("Poll not found.");
+            }
+
+            var now = DateTime.UtcNow;
+            poll.Title = dto.Title.Trim();
```
edgarasLegusVisma, line 201: Codex 

Renaming a poll leaves its wall with the old name
When a poll is renamed, could we update its private wall’s name and description as well? Currently only the Poll entity is updated, so conversation notifications can continue using the old poll title.
Reproduction:
1. Create a poll called “Old title”.
2. Have another user vote.
3. Rename the poll to “New title”.
4. Add a conversation post.
5. The notification/email still refers to “Old title”.

src/api/Shrooms.Presentation.Api/Controllers/PollController.cs
arturasvisma: Fixed, the soft deleted poll kept a tracked WallId so EF threw when the wall was removed, the Poll to Wall relationship is now ClientNoAction and there is a regression test.

src/api/Shrooms.Domain/Services/Polls/PollService.cs
arturasvisma: True, but events and projects create their wall the same way, so leaving it as is for now.

src/api/Shrooms.Domain/Services/Polls/PollService.cs
arturasvisma: Fixed, it now compares the submitted ballot with the stored one and rejects any change.

src/api/Shrooms.Domain/Services/Polls/PollService.cs
arturasvisma: Fixed, it now throws instead of adding a day.

IgnasJank (approved):

src/api/Shrooms.Domain/Services/Polls/PollService.cs
arturasvisma: Fixed, the wall name and description now follow the poll title on update, not pushed yet.

arturasvisma:
1. Fixed, the join now runs before the save.
2. Fixed by skipping the join for anonymous polls, seeding the whole audience would spam every user with every poll comment.
3. Not fixed, events and projects create their wall the same way, so this needs a shared transaction fix.
4. Fixed, it now compares the submitted ballot with the stored one and rejects any change.
5. Fixed, it now matches only a unique index violation and rethrows the rest.
6. Not fixed yet, agreed it should be a 404 with a clean body.
7. Six tests added for delete, voting, anonymity and rename, still thin on the state machine and canSeeResults.

Lower priority: publish now throws on a past deadline, and a title edit now updates the wall name and description. Paging, the validation order and the permission seeding are still open.

src/api/Shrooms.Domain/Services/Polls/PollService.cs
arturasvisma: Pushed now in 4565e7b2.

edgarasLegusVisma (approved):

PR #340, opened 2026-09-08
feat(events): own the recurring event schedule in Hangfire

IgnasJank:
New occurrences only appeared when an external scheduler POSTed /externalpremiumjobs/updaterecurringevents, and nothing in the solution registered that schedule. Hangfire now owns it, running weekdays 08-16 every one hour in the default organization timezone. The endpoint stays as a manual trigger, and the job is guarded against overlapping runs that would each clone the same expired occurrence.

The rollover now shifts dates until the occurrence lands in the future, so one run recovers from a missed schedule instead of crawling forward a period at a time, and it carries the RSVP and upcoming-widget flags that were previously dropped from every new occurrence.


arturasvisma:
Same bug, other fix — #343 makes `FLE Standard Time` resolve on Linux so `LocalStartDate` stops throwing, while this PR still calls it in the new rollover loop and only catches at schedule setup, where it falls back to UTC.

PR #342, opened 2026-09-10
Fix/service request search

IgnasJank:
Searching by name and last name was not working properly since these two are in different DB tables.


arturasvisma:
I do not understand, it is supposed to be changes regarding `service request search`, but most of the change is about events?

IgnasJank:
> I do not understand, it is supposed to be changes regarding `service request search`, but most of the change is about events?

Fixed. contained changes from other branch. My bad.

arturasvisma (approved):

PR #344, opened 2026-09-16
feat(groups): award monthly group kudos from a Logic App

arturasvisma:
Group types that carry a kudos type now pay their members automatically, instead of a kudos admin doing it by hand each month.

## How it runs

A Logic App calls `POST /ExternalPremiumJobs/AwardMonthlyGroupKudos` on the 1st of the month — basic auth plus the `Organization` header, same as `GiveLoyaltyKudos` and the other scheduled jobs. It awards **the month that just ended**, so a membership that started or ended late in that month is already settled.

`?year=&month=` awards that single month instead, for manual repair. Passing only one of the two is rejected rather than silently treated as a catch-up.

Recurrence for the workflow — `monthDays: [1]`, UTC:

```json
{ "frequency": "Month", "interval": 1, "timeZone": "UTC", "schedule": { "monthDays": [1], "hours": [2], "minutes": [0] } }
```

## Wording

`GroupType.AwardTemplate` (nvarchar 500) holds each type's own wording, with `{groupname}`, `{role}` and `{month}`; several groups or roles join with ` | `. `{role}` is the member's `GroupMember.Description`. Empty falls back to `Monthly group kudos for {month}: {groupname}`.

```
For your valuable contribution to the Ethics Committee (2026-08)! 🏅🙌👏
Foodmaster (2026-08) Role3 | Role1 🍕
Monthly group kudos for 2026-08: Carpool Coordinators
```

Allocations group by **group type**, not kudos type, so two types sharing one kudos type each get their own log with their own wording.

## Not awarding twice

`KudosLog.GroupKudosPeriod` (nullable, first day of the month, filtered index) records which month a log paid for. A retry finds it and reports `alreadyAwarded`. The walk collects every unawarded month rather than stopping at the first awarded one, so a month awarded on its own leaves no gap that could never be filled.

The catch-up is capped at **3 months**. Without the cap, a group type given a kudos type long after the 2026-07 floor would back-pay every month since it — open-ended memberships count as active in all of them. The cap also lets a month that allocated nothing fall out of the window instead of being rescanned forever.

## Who pays whom

Excluded: temporary types (they pay once at the end of their term), groups still awaiting approval, memberships not overlapping the month. Several memberships of one group pay once but contribute every role.

## Verified against a real database

Seeded eight users across committee, food master, temporary and untemplated types with varied start and end dates, then ran the endpoint:

| Case | Result |
|---|---|
| Explicit 2026-08 from clean | 6 logs / 60 points |
| Retry | `alreadyAwarded`, nothing written |
| Scheduled path, August done, July missing | July only — gap filled |
| Scheduled path again | `[]` |
| Scheduled path from scratch | both months, neither skipped |
| Scheduled path, no memberships | nothing written |

Slowest call 0.34s, the rest tens of milliseconds.

Two bugs were caught this way and are fixed here: the catch-up used to skip the newest month because its own writes looked like a prior award, and the 3-month cap came out of reviewing the back-pay behaviour.

## Tests

`dotnet test Shrooms.sln` — 1624 passed. The 2 failures in `Shrooms.Premium.Tests` are pre-existing locale assertions in the event Excel export (`"First name"` vs `"Vardas"`), confirmed failing on a clean `master`.

## Frontend

The admin field for the template is SimoonaV2/simoona-nextjs#368.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


IgnasJank:
Claude findings

> 
> 1. Groups/AwardMonthlyKudos was left behind and is now a trap — GroupsController.cs:161-185
> 
> Its XML doc still reads "Logic App entry point. Not idempotent — the caller owns scheduling, so running it twice for the same period awards twice." Both halves are now false: the Logic App entry point moved to ExternalPremiumJobs/AwardMonthlyGroupKudos, and AwardMonthlyKudosAsync short-circuits on AlreadyAwarded.
> 
> The real problem is the default. That endpoint still defaults to the current month (month ?? now.Month, line 179), while the webhook defaults to the previous one. Now that every award stamps GroupKudosPeriod, a kudos admin who calls it on the 12th pays only the members active so far and permanently marks the month awarded — the scheduled run on the 1st reports AlreadyAwarded and nobody else ever gets paid for that month. Nothing in the frontend calls this endpoint. Deleting it is the cleanest fix; otherwise default it to the previous month and rewrite the doc.
> 
> 2. The idempotency check is read-then-write with nothing guarding it — GroupKudosService.cs:150, 227
> 
> IsAlreadyAwardedAsync reads, then AwardPeriodAsync writes, and the new IX_KudosLogs_OrganizationId_GroupKudosPeriod is non-unique so the database won't catch a duplicate. Two overlapping calls — a Logic App retry on a slow response, or a retry racing the admin endpoint above — both see the period unawarded and both write a full batch of pending logs. LoyaltyKudosService in the same folder guards exactly this shape with a static SemaphoreSlim; mirroring that would close it.
> 
> 3. A rejected month can never be re-run — GroupKudosService.cs:227-234
> 
> The check asks only whether a stamped log exists, not what became of it. If a run goes out with a bad template or a wrong membership snapshot and the kudos admin rejects the batch, the month stays "awarded" forever — the only recovery is editing GroupKudosPeriod by hand in the database. Excluding KudosStatus.Rejected from that AnyAsync would make a rejection mean "try again", which is what rejecting a batch normally means.
> 
> Smaller notes
> 
> - LogOutcome says "nothing outstanding" (GroupKudosWebHookService.cs:80) when a month has just aged out past MaxCatchUpMonths and become unpayable. The 3-month window is clearly deliberate and well covered by tests, but a month falling off the end deserves a warning rather than silence.
> - {role} with no role descriptions renders an empty slot: "Foodmaster {role} 🍕" → "Foodmaster  🍕". Not whitespace-only, so the empty-render fallback doesn't catch it.
> - I checked the CreatedBy = null path — BasicAuthValidator issues no NameIdentifier claim, so UpdateEntityMetadata leaves it null and KudosLog.Approve's userId == CreatedBy guard can't misfire. Fine as-is.

IgnasJank (approved):

PR #345, opened 2026-09-17
feat(kudos): return reactions with a user's kudos log

edgarasLegusVisma:
## What

`GetUserKudosLogs` now returns `likes` on every entry, the same shape the wall widget's `GetLastKudosLogsForWall` already carries. The new UI's kudos log needs it to show reactions

## How

- `Likes` added to `KudosUserLogDto` and `KudosUserLogViewModel`; the existing `LikeDto → LikeViewModel` map covers the conversion.
- `GetUserKudosLogsAsync` loads the `Likes` column for the page's log ids in a second, `AsNoTracking` query (EF Core refuses to project an owned collection from a tracking one) and resolves liker names through `GetLikersByIdAsync`, which the wall query now shares.
- Users that no longer exist are dropped from the list, as on the wall.


IgnasJank:
Approved. Claude findings a smaller. 


1. Medium-High — the new owned-type SQL projection is the codebase's first, and nothing tests it KudosService.cs:291

.AsNoTracking()
.Where(log => logIds.Contains(log.Id))
.Select(log => new { log.Id, log.Likes })

KudosLog.Likes is an owned type sharing the owner's table (OtherEntitiesConfig.cs:68, OwnsOne → the Likes JSON column). Every other read of .Likes in the codebase runs against already-materialized entities — GetLastKudosLogsForWallAsync:336 and WallService.GetUsersAsync:793 both take a List/IReadOnlyCollection and do LINQ-to-Objects. This is the only place an owned entity gets projected into SQL.

And the new test can't catch it: KudosServiceTests uses NSubstitute DbSet mocks backed by a plain in-memory IQueryable provider (MockingExtensions.cs:95-107), so EF is never asked to translate anything. AsNoTracking() is likewise a no-op there — it returns source unchanged when the provider isn't an EntityQueryProvider — so the tracking-vs-no-tracking distinction that makes this projection legal in the first place is also unverified.

Failure mode: if EF Core 10 refuses the projection, GET /Kudos/GetUserKudosLogs throws on every call while CI stays green. Either add one integration test against a real provider, or fold the likes into the main query (below) and sidestep the question.

2. Medium — the second round trip is avoidable KudosService.cs:285-298

userLogsQuery already selects from kudLog; re-querying _kudosLogsDbSet by the ids it just returned costs an extra round trip and a 50-parameter IN clause per page. Projecting into new { Dto = new KudosUserLogDto { … }, Likes = kudLog.Likes } and mapping after materialization gets it in one query and drops logIds, the dictionary and the GetValueOrDefault lookups. Minor note: the second query also isn't scoped by OrganizationId — harmless today because the ids come from the scoped first query, but it loses the defence-in-depth the rest of the method has.

3. Medium — reactions now reach row kinds the wall deliberately excludes KudosService.cs:244, 363-374

GetUserKudosLogsAsync filters only on EmployeeId and OrganizationId, so Minus, Refund, basket donation, welcome and automatic rows all come back with a Likes array — while GetLastKudosLogsForWallAsync:328 explicitly excludes Minus/Refund. ToggleLikeAsync gates on Approved and org only, so those rows are genuinely reactable end to end now. Since the sender's Minus row is a separate KudosLog from the receiver's Send row, a reaction left there lands on a log id no other surface ever renders. This is the backend half of finding 2 from the frontend review — fixing it in ToggleLikeAsync (mirror the wall's type exclusion) covers both.

4. Low — a process-wide lock newly under real load KudosService.cs:40

_kudosLogLikeLock is a single static SemaphoreSlim(1, 1) held across await _uow.SaveChangesAsync, serializing every kudos reaction toggle across all users and all organizations in the process. Until now the only reaction surface was the 5-item wall widget (WallKudosLogCount = 5); the user log adds up to 50 reactable rows per page view. Pre-existing code, but this change is what makes the contention matter.

IgnasJank (approved):

